PR #93, opened 2016-02-01
updated constraint-declaration-validation.xml to contain Default grou…

lpandzic:
Continuation of https://hibernate.atlassian.net/browse/BVAL-493
I'm not sure if anything more should be updated here?


gunnarmorling:
Going to close this in favor of #234. Thanks again for the suggestion, @lpandzic!

PR #95, opened 2016-02-12
Add scripts to publish the latest draft on the websites

DavideD:
Added the scripts to sync the latest draft on beanvalidation.org

@Sanne  can you check that they are still compatible with the things you are doing?
@emmanuelbernard should we keep them on master or should we create a production/staging branch.
I would keep them on master.


gunnarmorling:
Closing, I'll send a new one shortly. Thanks, @DavideD.

PR #100, opened 2016-08-18
BVAL-527, BVAL-529 Setting baseline to Java 8; making apparent that BV 2.0 is current

gunnarmorling:


gunnarmorling:
@gsmet There you go :)

PR #101, opened 2016-08-18
BVAL-530 Rename master.asciidoc to index.asciidoc

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-530

It is not possible to change the name of the output file and we want it
to be named index.html.

As for the PDF file, we would have needed to rename it anyway if we want
to publish it.


gunnarmorling:
Thanks, @gsmet!

PR #102, opened 2016-08-22
BVAL-455 Updating spec for 2.0 XML schemas

gunnarmorling:
https://hibernate.atlassian.net/browse/BVAL-455

### DIFF
--- a/sources/xml-descriptor.asciidoc
+++ b/sources/xml-descriptor.asciidoc
@@ -44,11 +44,11 @@ The `ignore-annotations` setting is not inherited from nor by the class hierarch
 ----
 <?xml version="1.0" encoding="UTF-8"?>
 <constraint-mappings
-        xmlns="http://jboss.org/xml/ns/javax/validation/mapping"
+        xmlns="http://xmlns.jcp.org/xml/ns/validation/mapping"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation=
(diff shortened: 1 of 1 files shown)


hendrikebbers:
Until now we do not have a schema at "http://xmlns.jcp.org/xml/ns/validation/mapping validation-mapping-2.0.xsd". I have no idea how complex it will be to upload such a file to the jcp domain. As said I think that the change is a good one but maybe we should check that we can easily upload new xsd files before we change it.

gunnarmorling:
Hey @hendrikebbers, thanks for the feedback!

Yes, we need to work with the folks at JCP how to get the files there; there is [BVAL-531](https://hibernate.atlassian.net/browse/BVAL-531) for tracking this. I don't expect big problems given that the other schema files are hosted there, too. I'll ask the PMO what the process is, I suppose it may take some time to make it happen.

hferentschik:
+1 for the change, but let's double check with the process of getting this files hosted first.

gunnarmorling:
I've sent a mail to the PMO, asking about the process. But we can move
forward with the change for now IMO, as we can live without the files
exposed through that URI for the time being and then have the final
versions hosted once we don't expect any further changes (nothing really
mandates that schema files are available through the URI given in the
namespace, it's just a common expectation people have).

hendrikebbers:
+1
Uploading the final version should be fine

Von meinem iPhone gesendet

gsmet:
Amended and merged. Thanks!

PR #103, opened 2016-08-22
BVAL-534 Update all outdated links to java.sun.com

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-534


gunnarmorling:
Merged; thanks, @gsmet. Btw. I think we can move to the simpler link syntax, it's more readable in the source. No need to bulk-change everything, but when adding new stuff or touch existing links anyways, let's use the simpler one.

PR #109, opened 2016-08-27
BVAL-540 Use Maven to download the snapshot of the API sources

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-540
 * https://hibernate.atlassian.net/browse/BVAL-538


gunnarmorling:
Merged; thx, @gsmet!

PR #110, opened 2016-08-29
BVAL-497 Take into account @Repeatable annotations in the spec

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-497


gunnarmorling:
I've sent PR #111 based on this, thus closing this one. The wording in the spec is a bit different, I also pushed out the example into separate source files. @gsmet if it's good from your PoV, I'll refer the EG to the imminent change.

PR #111, opened 2016-08-30
BVAL-497 Marking built-in constraints with @Repeatable

gunnarmorling:


gsmet:
@gunnarmorling Looks good to me.

sources/constraint-definition.asciidoc
```diff
@@ -490,64 +490,37 @@ It is often useful to declare the same constraint more than once to the same tar
...
 ====
 
 In this example, both constraints apply to the [methodname]`zipcode` field but with different groups and with different error messages.
+It is also possible to specify a constraint several types by explicitly using the [classname]`@List` annotation (though simply repeating the annotation is the preferred idiom as of Bean Validation 2.0 and Java 8):
+
+.Multi-valued constraint declaration using explicit [classname]`@List` annotation
+====
+[source, JAVA, indent=0]
+----
+include::{spec-examples-source-dir}org/beanvalidation/specexamples/chapter03/multivaluedconstraint/explicitlist/Address.java[tags=include]
```
gsmet: Is it such a good idea to use the chapter number? What if we add a new chapter?

I think it would probably be better to use the name of the .adoc file as the first directory.

gunnarmorling:
@gsmet I've pushed an update as per your last comment.

gsmet:
Amended and merged! Thanks!

PR #112, opened 2016-09-14
BVAL-544 Promote TYPE_USE usage for constraint annotations

gsmet:


gunnarmorling:
It's a bit unfortunate that the commit mixes the extraction of code into source files and the actual changes. We should keep these things separate for the sake of allowing people more easily understand what actual has changed. Do you think you could rework the commit to split this up? I'd normally say "let's do it better next time", but in case of the spec repo I'd prefer to have this properly done. Thanks!

gsmet:
Not convinced it was time well spent but I updated the PR.

gunnarmorling:
Thanks for doing the update, @gsmet. Rebased and applied.

PR #114, opened 2016-09-23
BVAL-498 Mandating retrieval of parameter names from class files if p…

gunnarmorling:
…resent

A simple one as it's Friday :)

### DIFF
--- a/sources/constraint-declaration-validation.asciidoc
+++ b/sources/constraint-declaration-validation.asciidoc
@@ -836,7 +836,7 @@ Some constraints can target an executable's return value as well as its array of
 
 ===== Naming parameters
 
-In case the validation of a parameter constraint fails, the concerned parameter needs to be identified in the resulting [classname]`ConstraintViolation` (see section <<validationapi-constraintviolation>>). As of version 7, Java doesn't provide a portable way to retrieve parameter names. Bean Validation therefore defines the [classname]`javax.validation.ParameterNameProvider` API to which the retrieval of parameter names is delegated:
+In case the validation of a parameter constraint fails, the concerned parameter needs to be identified in the resulting [classname]`ConstraintViolation` (see section <<validationapi-constraintviolation>>). Bean Validation defines the [classname]`javax.validation.ParameterNameProvider` API to which the retrieval of parameter names is delegated:
 
 
 [source, JAVA]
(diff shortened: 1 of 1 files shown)


sources/constraint-declaration-validation.asciidoc
```diff
@@ -876,7 +876,12 @@ public interface ParameterNameProvider {
 }
 ----
 
-[tck-testable]#A conforming Bean Validation implementation provides a default [classname]`ParameterNameProvider` implementation which returns parameter names in the form arg``PARAMETER_INDEX``, where `PARAMETER_INDEX` starts at 0 for the first parameter, e.g. [varname]`arg0`, [varname]`arg1` etc.#
+[tck-testable]#A conforming Bean Validation implementation provides a default [classname]`ParameterNameProvider` implementation which returns parameter names as stored in the class file containing the valided executable, if present. A conforming implementation must either use the Java reflection API or ensure behavioral compatability to using the reflection API in this way:
```
gsmet: Hmmm, I nearly fall into your trap! Looks like [tck-testable]# does not support being spanned on multiple lines like that. I would have fixed it on merging but I'm not sure of the best way to fix that as in this case, the items in the list are the important part.

Btw, s/compatability/compatibility/ (even if there is an entry in the wiktionary for compatability: https://en.wiktionary.org/wiki/compatability)

gunnarmorling: Hum, good catch with the mark-up. What would end up in the assertion file right now?

gunnarmorling:
Force-pushed an update to fix the typo.

sources/constraint-declaration-validation.asciidoc
gunnarmorling: @gsmet I've force-pushed an update which marks only the first sentence as "tck-testable". It's not optimal, but the trailing ":" makes clear that there is more to come which is part of the assertion. Lacking a better solution I think it's acceptable. Can you merge it?

gsmet: FYI, the main issue was that the `[tck-testable]#` appeared in clear in the spec as it wasn't replaced by our custom block processor.

gunnarmorling:
I see. If there is any way to add the list contents to the assertion,
please feel free to change it :)

gsmet:
Fixed a remaining typo and merged! Thanks. I'll work on the TCK adjustments this afternoon.

PR #115, opened 2016-10-06
BVAL-498 Include the bullet points in the tck-audit.xml file

gsmet:
@gunnarmorling too simple to be true :)


gunnarmorling:
Nice one; didn't we have a similar problem somewhere else?

gsmet:
Maybe, but it's not easy to find them. I looked for `:</text>` without success. So we should fix them when we see them.

PR #116, opened 2016-12-22
BVAL-508 Adding latest proposal for value extraction as an appendix

gunnarmorling:
@emmanuelbernard That's the latest proposal around extractors, based on our recent discussions and insights from the PoC in the RI. It takes an opinion on some of the questions from the original proposal. Some remaining questions are listed at the end. Overall I think we are getting there.

### DIFF
--- a/sources/appendix-value-extraction.asciidoc
+++ b/sources/appendix-value-extraction.asciidoc
@@ -5,32 +5,32 @@
 
 [[appendix-value-extraction]]
 
-== Value extraction for cascaded validation and type-level constraints (BVAL-508)
+== Value extraction for cascaded validation and type argument constraints (BVAL-508)
 
-This appendix describes the current work-in-progress around the retrieval of values to be validated during cascaded validation and evaluation of type-level constraints. It is based on the original http://beanvalidation.org/proposals/BVAL-508/[proposals for BVAL-508].
+This appendix describes the current work-in-progress around the retrieval of values to be validated during cascaded validation and evaluation of type argument constraints. It is based on the original http://beanvalidation.org/proposals/BVAL-508/[proposals for BVAL-508].
 
 === Motivation
 
 Value extraction is needed when constraints are applied to the element(s) stored within a container type. There are two categories:
 
-* Cascaded validation of collections, maps and arrays as triggered via `@Valid`: `@Valid List<Order> orders`; in this case all the values stored in the list must be extracted so each can be validated
-* Validation of type-level constraints as enabled by Java 8: `Optional<@Email String> email` or `Property<@Min(1) Integer> value`; in this case, e.g. the wrapped `String` and `Integer` values must be extracted so the `@Email` and `@Min` constraints can be applied
+* Cascaded validation of iterables, maps and arrays as triggered via `@Valid`: `@Valid List<Order> orders`; in this case all the values stored in the list must be extracted so each can be validated
+* Validation of type argument constraints as enabled by Java 8: `Optional<@Email String> email` or `Property<@Min(1) Integer> value`; in this case, e.g. the wrapped `String` and `Integer` values must be extracted so the `@Email` and `@Min` constraints can be applied
 
-Both cases can also overlap: `List<@RetailOrder @Valid> retailOrders`.
+Both cases can also overlap: `List<@RetailOrder @Valid Order> retailOrders`.
 
-Type-level constraints and pluggable extractors will also make cascaded validation more flexible. As of BV 1.1, there was a fixed set of data types supported by cascaded validation mandated by the spec (`Iterable`, arrays, `Map` (only the values would be validated)). This excludes use cases such as custom collection types (e.g. Guava's multi-map), validation of map keys or collection types of other JVM languages (such as Ceylon's collection framework).
+Type argument constraints and pluggable extractors will also make cascaded validation more flexible. As of BV 1.1, there was a fixed set of data types supported by cascaded validation mandated by the spec (`Iterable`, arrays, `Map` (only the values would be validated). This excludes use cases such as custom collection types (e.g. Guava's multi-map), validation of map keys or collection types of other JVM languages (such as Ceylon's collection framework).
 
 Allowing to put `@Valid` on type parameters allows to express the subject of cascaded validation more specifically: `Map<@Valid AddressType, @Valid Address> addresses` would describe that both, map keys and values, should be validated.
 
 === Requirements
 
-1. Constraints can be applied to type parameters (of bean properties, but also method parameters etc.): `Property<@Min(1) Integer> value`
+1. Constraints can be applied to type arguments (of bean properties, but also executable parameters and method return values): `Property<@Min(1) Integer> value`
 1. Individual type parameters can be marked for cascaded validation: `Map<@Valid AddressType, Address> addresses`
-1. How values are retrieved must be customizable by means of a value extractor SPI, with a set of defined default extractors
+1. How values are retrieved must be customizable by means of a pluggable value extractor mechanism, with a set of defined default extractors
 1. Sometimes constraints should apply to a value wrapped by a container type, but there is no type parameter to put the constraints to. JavaFX's `Property` hierarchy is the most prominent example: `@Email StringProperty emailProperty`. In such case the value should be implicitly extracted from the container if unambiguously doable.
 1. OPEN: Support for validation of nested collections: `List<Map<String, @NotNull String>> addressesByType`.
-1. Allow to configure type-level constraints and cascades via XML mappings
-1. Expose meta-data on type-level constraints and cascades in the constraint metadata API
+1. Allow to configure type argument constraints and cascades via XML mappings
+1. Expose meta-data on type argument constraints and cascades in the constraint metadata API
 
 === Non-requirements
 
@@ -45,7 +45,7 @@ class CustList extends List<@NotNull Customer> {
 
 === Solution
 
-Value retrieval for cascaded validation and validation of type-level constraints is done via the `ValueExtractor` SPI:
+Value retrieval for cascaded validation and validation of type argument constraints is done via the `ValueExtractor` API:
 
 [source,java]
 ----
@@ -78,10 +78,8 @@ class ListValueExtractor implements ValueExtractor<List<@ExtractedValue ?>> {
 
 	@Override
 	public void extractValues(List<?> originalValue, ValueReceiver receiver) {
-		int i = 0;
-		for ( Object object : originalValue ) {
-			receiver.indexedValue( object, "<collection element>", i );
-			i++;
+		for ( int i 0 = 0; i < originalValue.length; i++ ) {
+			receiver.indexedValue( originalValue.get( i ), "<iterable element>", i );
 		}
 	}
 }
@@ -91,6 +89,47 @@ The right callback methods (`indexedValue`, `keyedValue` etc.) must be called in
 
 If a non-null value is passed for `nodeName`, a path node of type `TYPE_PARAMETER` will be appended to the property path. That's desirable for collection types for instance. If null is passed, no node will be appended, resulting in the same path as if the constraint had been given on the element itself instead of a type parameter. That's desirable for pure "wrapper types" such as `Optional`.
 
+==== @ExtractedValue
+
+The `@ExtractedValue` annotation is used to denote the element extracted by a given value extractor:
+
+[source,java]
+----
+package javax.validation.valueextraction;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * @author Gunnar Morling
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE_USE)
+public @interface ExtractedValue {
+}
+----
+
+The `@ExtractedValue` annotation must be specified exactly once for a value extractor type. Thus the following extractor definition is illegal as it specifies `@ExtractedValue` several times:
+
+[source,java]
+----
+public class DoubleExtractor implements ValueExtractor<Multimap<@ExtractedValue ?, @ExtractedValue ?>> { ... }
+----
+
+When defined for a generic type, only wildcard type arguments may be annotated with `@ExtractedValue`. I.e. the following extractor definition is illegal:
+
+[source,java]
+----
+public class StringListValueExtractor implements ValueExtractor<List<@ExtractedValue String>> { ... }
+----
+
+[NOTE]
+====
+This implies that there may not be more than one extractor for a given generic type. I.e. there can be an extractor for `List<?>`, but not one for `List<String>` and one for `List<Integer>`.
+====
+
 In some cases, an extractor applies to the entire extracted type itself (e.g. for arrays). In this case the `@ExtractedValue` annotation is to be given on the extracted type itself:
 
 [source,java]
@@ -99,10 +138,8 @@ class ObjectArrayValueExtractor implements ValueExtractor<@ExtractedValue Object
 
 	@Override
 	public void extractValues(Object[] originalValue, ValueReceiver receiver) {
-		int i = 0;
-		for ( Object object : originalValue ) {
-			receiver.indexedValue( object, "<collection element>", i );
-			i++;
+		for ( int i = 0; i < originalValue.length; i++ ) {
+			receiver.indexedValue( originalValue[i], "<iterable element>", i );
 		}
 	}
 }
@@ -116,20 +153,59 @@ Instead of returning the extracted values from the method call, implementations 
 
 ==== Default extractors
 
-Compatible implementations provide extractors for the following types out of the box:
+Compatible implementations provide extractors for the following types, applying the described semantics, out of the box:
 
-* arrays of objects and all primitive data types
-* `java.util.Iterable` (when `@Valid` is given on the iterable element itself, the element and all its entries will be validated; this is to grant backwards compatability with BV 1.1; when `@Valid` is given on the type parameter of an iterable element, all the entries will be validated). A node of type `TYPE_PARAMETER` and with the name "<collection element>" will be added to the path when validating a type parameter constraint.
-* `java.util.Map` (when `@Valid` is given on the map element itself, the element and all its values will be validated; this is to grant backwards compatability with BV 1.1; when `@Valid` is given on the key type parameter, only the keys will be validated; when `@Valid` is given on the value type parameter, only the values will be validated. A node of type `TYPE_PARAMETER` and with the name "<map value>" will be added to the path when validating a type parameter constraint on the map value. A node of type `TYPE_PARAMETER` and with the name "<map key>" will be added to the path when validating a type parameter constraint on the map key.
-TODO: format of key representations
-* `java.util.Optional`. No node will be appended to the path when validating type-level constraints on `Optional`. Constraints given on an element of type `Optional` apply to the wrapped value by default (see below).
-* `javafx.beans.observable.ObservableValue`. No node will be appended to the path when validating type-level constraints on `ObservableValue`. Constraints given on an element of type `ObservableValue` apply to the wrapped value by default (see below).
+* Arrays of objects and all primitive data types
+** `@Valid` can be given for the array itself or for its component type, which will cause the validation of all the array elements
+** If a constraint given for an array's component type is validated, a node with the following properties will be added to the path:
+*** name: "<iterable element>"
+*** kind: `TYPE_PARAMETER`
+*** isInIterable: `false`
+*** index: the element's index
+*** key: `null`
+* `java.util.Iterable`
+** When `@Valid` is given on the iterable element itself, the element and all its entries will be validated; this is to grant backwards compatability with BV 1.1
+** When `@Valid` is given on the type parameter of an iterable element, all the entries will be validated).
+** When validating a type argument constraint for `Iterable`, a node with the following properties will be added to the path:
+*** name: "<iterable element>"
+*** kind: `TYPE_PARAMETER`
+*** isInIterable: `true`
+*** index: the element's index if the iterable is of type `List` or a subtype thereof; `null` otherwise
+*** key: `null`
+* `java.util.Map`
+** When `@Valid` is given on the map element itself, the element and all its values will be validated; this is to grant backwards compatability with BV 1.1
+** When `@Valid` is given on the key type parameter, only the keys will be validated
+** When `@Valid` is given on the value type parameter, only the values will be validated
+** When validating a constraint on the key type argument of `Map`, a node with the following properties will be added to the path:
+*** name: "<map key>"
+*** kind: `TYPE_PARAMETER`
+*** isInIterable: `true`
+*** index: `null`
+*** key: "(" + key + ")"; TODO: re-consider
+** When validating a constraint on the value type argument of `Map`, a node with the following properties will be added to the path:
+*** name: "<map value>"
+*** kind: `TYPE_PARAMETER`
+*** isInIterable: `true`
+*** index: `null`
+*** key: key; TODO: re-consider
+* `java.util.Optional`
+** No node will be appended to the path when validating type argument constraints on `Optional`
+** Constraints given on an element of type `Optional` apply to the wrapped value by default (see <<appendixvalueextraction-wrappedelements>>).
+* `javafx.beans.observable.ObservableValue`
+** No node will be appended to the path when validating type argument constraints on `ObservableValue`
+** Constraints given on an element of type `ObservableValue` apply to the wrapped value by default (see <<appendixvalueextraction-wrappedelements>>).
+* `java.lang.Object`
+** When `@Valid` is given for an element, the element will be validated
+
+===== Examples
+
+TODO
 
 ==== Plugging in custom extractors
 
 Additional value extractors can be registered with the Bean Validation provider in the following ways:
 
-* Invoke the new method `Configuration#addValueExtractor(ValueExtractor<?>)` (to apply it on the validation factory level)
+* Invoke the new method `Configuration#addValueExtractor(ValueExtractor<?>)` (to apply it at the validation factory level)
 * Invoke the new method `ValidatorContext#addValueExtractor(ValueExtractor<?>)` (to apply it for a single `Validator`)
 * Specify the fully qualified class name of one or several extractors in `META-INF/validation.xml`: 
 
@@ -142,7 +218,7 @@ Additional value extractors can be registered with the Bean Validation provider 
 
 ==== Retrieval of extractors
 
-When detecting a type-level constraint or cascade, the applicable extractor is determined as follows:
+When detecting a type argument constraint or cascade or a non type argument constraint that is subject to value extraction, the applicable extractor is determined as follows:
 
 * Choose the most specific extractor matching the type parameter in question; an extractor A is more specific than another extractor B if A extracts a subtype of the type extracted by B.
 * If there are several extractors which are equally specific, an `UnexpectedTypeException` is raised. TODO: apply rules similar to "ConstraintValidator resolution algorithm".
@@ -155,7 +231,8 @@ When detecting a type-level constraint or cascade, the applicable extractor is d
 As extractor retrieval for type parameter constraints is done using the static type of constrained elements, the retrieval can be done once at initialization time and then be cached. This is not possible for retrieval of extractors for cascaded validation.
 ====
 
-==== Applying element-level constraints to wrapper elements
+[[appendixvalueextraction-wrappedelements]]
+==== Applying element-level constraints to wrapped elements
 
 Sometimes there is no type parameter to put a constraint to, but still constraints should be applied to the wrapped value instead of the annotated element. JavaFX's property hierachy falls into that category, as it defines specific `Property` sub-types which are not generic:
 
@@ -191,7 +268,7 @@ Value extractor definitions can be marked with `ConstraintsApplyTo` so that cons
 class PropertyExtractor implements ValueExtractor<Property<@ExtractedValue>> {
 
 	@Override
-	public void extractValues(Property[] originalValue, ValueReceiver receiver) {
+	public void extractValues(Property originalValue, ValueReceiver receiver) {
 		// ...
 	}
 }
@@ -211,6 +288,25 @@ In rare cases it may happen that a constraint should be applied to the wrapped v
 .Limitations
 ====
 When looking for an extractor in this case, only the extracted types can be considered (e.g. `Map`), but not the extracted type parameters (e.g. Map's `K` or `V` parameters). If there are multiple extractors for the same most specific type (e.g. the `Map` key and value extractors), no extractor can be chosen and an exception will be raised. I.e. automatic unwrapping is only supported for cases where the super-type in question has a single type parameter.
+====
+
+[NOTE]
+====
+For the sake of readability, when applying constraints to the elements of a generic container type, it is strongly recommended to put the constraints to the type argument instead of the element itself in conjunction with `@ConstraintsApplyTo`. I.e. you should prefer
+
+[source,java]
+----
+List<@Email String> emails;
+----
+
+over
+
+[source,java]
+----
+@Email
+@ConstraintsApplyTo(WRAPPED_VALUE)
+List<String> emails;
+----
 ====
 
 ==== Examples
@@ -238,6 +334,8 @@ List<@NotNull @Email String> emails;
 List<@Valid Order> orders;
 ----
 
+This will validate the constraints on each `Order` element in the list.
+
 * The legacy style for cascaded validation is supported as well:
 
 [source,java]
@@ -247,7 +345,7 @@ List<@Valid Order> orders;
 
 This would also validate any constraints on a custom list type (e.g. `MyList#getId()`). TODO: we never clarified that in 1.1. Should it be made explicit?
 
-* Map validation with type-level constraints and cascading:
+* Map validation with type argument constraints and cascading:
 
 [source,java]
 ----
@@ -287,7 +385,7 @@ private NumericMap<@Min(1) Integer> integerMap;
 
 The `@Min` constraint is to be applied to the map's keys and values as the annotated type parameter maps to `K` and `V` of `Map`.
 
-* Type-level constraints can be applied to the elements of `Object` arrays and arrays of any primitive type:
+* type argument constraints can be applied to the elements of `Object` arrays and arrays of any primitive type:
 
 [source,java]
 ----
@@ -338,7 +436,6 @@ public class TupleFirstExtractor implements ValueExtractor<Tuple<@ExtractedValue
 	@Override
 	public void extractValues(Tuple<?, ?> originalValue, ValueReceiver receiver) {
 		receiver.value( originalValue.getFirst(), "<first>" );
-		}
 	}
 }
 ----
@@ -350,7 +447,6 @@ public class TupleSecondExtractor implements ValueExtractor<Tuple<?, @ExtractedV
 	@Override
 	public void extractValues(Tuple<?, ?> originalValue, ValueReceiver receiver) {
 		receiver.value( originalValue.getSecond(), "<second>" );
-		}
 	}
 }
 ----
@@ -361,13 +457,6 @@ private Tuple<@NotNull @Email String, @NotNull @Min(1) Integer> tuple;
 ----
 
 ==== Invalid examples
-
-* Extractor types must declare exactly one parameter as the extracted value. Hence this extractor type is invalid:
-
-[source,java]
-----
-public class DoubleExtractor implements ValueExtractor<Multimap<@ExtractedValue ?, @ExtractedValue ?>> { ... }
-----
 
 * No most specific extractor can be found unambiguously:
 
@@ -417,14 +506,22 @@ TODO
 
 ==== Misc.
 
-* Regarding group sequences and default group sequences, the same rules apply for type-level constraints as they apply for regular element-level constraints.
+* Regarding group sequences and default group sequences, the same rules apply for type argument constraints as they apply for regular element-level constraints.
+* For the conversion of validation groups the same rules apply no matter whether `@Valid` is given for a regular element or for a type argument. I.e. the following group conversion declaration is valid:
+
+[source,java]
+----
+private List<@Valid @ConvertGroup(from=Default.class, to=Other.class) Order> orders;
+----
 
 === Open questions
 
 * `ConstraintsApplyTo` only allows one behavior per annotated element. Should it be per constraint? E.g. for `@NotNull @Email StringProperty email` it may be desirable to apply `@NotNull` to the wrapper but `@Email` to the wrapped value. That's not possible currently.
-* Should type-level constraints be validated when the container is null?
+* Should a path node be added for type argument constraints of `Optional` and similar types? This proposal suggests to not do it, but Emmanuel is not convinced of this.
+* Should type argument constraints be validated when the container is null?
+* Should value extractors be discoverable via the service loader mechanism (i.e. by means of `META-INF/services/javax.validation.valueextraction.ValueExtractor` files)
 * What to return from `PropertyDescriptor#getElementClass()` if there is a field of type `Foo` but a getter of type `Optional<Foo>`. So far, BV assumed the types of field and getter to be the same and exposed a single property descriptor (which btw. also may fall apart as of BV 1.1 when the field is of a sub-type of the getter's type). What to return here?
-* Should the presence of type-level constraints alone trigger cascaded validation? It doesn't seem right, but it may be useful when e.g. considering the case of `Tuple` above.
+* Should the presence of type argument constraints alone trigger cascaded validation? It doesn't seem right, but it may be useful when e.g. considering the case of `Tuple` above.
 * Should we allow extractors to be defined for specific parameterizations of generic types, e.g.
 
 [source,java]
@@ -435,3 +532,5 @@ public class ListOfStringExtractor implements ValueExtractor<List<@ExtractedValu
 ----
 
 I can't see a compelling use case for this (when would extractor behavior differ between different parameterizations of the same generic type) and am leaning towards only supporting the wildcard parameterization (`implements ValueExtractor<List<@ExtractedValue ?>>`).
+
+* Vet the API by exploring advanced use cases, e.g. Guava's https://github.com/google/guava/wiki/NewCollectionTypesExplained#table[Table], http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/graph/Graph.html[Graph], http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/graph/ValueGraph.html[ValueGraph] and http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/graph/Network.html[Network] types
--- a/sources/appendix-value-extraction.asciidoc
+++ b/sources/appendix-value-extraction.asciidoc
@@ -425,7 +425,7 @@ TODO
 * Should type-level constraints be validated when the container is null?
 * What to return from `PropertyDescriptor#getElementClass()` if there is a field of type `Foo` but a getter of type `Optional<Foo>`. So far, BV assumed the types of field and getter to be the same and exposed a single property descriptor (which btw. also may fall apart as of BV 1.1 when the field is of a sub-type of the getter's type). What to return here?
 * Should the presence of type-level constraints alone trigger cascaded validation? It doesn't seem right, but it may be useful when e.g. considering the case of `Tuple` above.
-* Should we allow extractors to be defined for specific paramterizations of generic types, e.g.
+* Should we allow extractors to be defined for specific parameterizations of generic types, e.g.
 
 [source,java]
 ----
@@ -434,4 +434,4 @@ public class ListOfIntegerExtractor implements ValueExtractor<List<@ExtractedVal
 public class ListOfStringExtractor implements ValueExtractor<List<@ExtractedValue String>> { ... }
 ----
 
-I can't see a compelling use case for this (when would extractor behavior differ between different parameterizations of the same generic type) and am leaing towards only supporting the wildcard parameterization (`implements ValueExtractor<List<@ExtractedValue ?>>`).
+I can't see a compelling use case for this (when would extractor behavior differ between different parameterizations of the same generic type) and am leaning towards only supporting the wildcard parameterization (`implements ValueExtractor<List<@ExtractedValue ?>>`).
--- a/sources/appendix-value-extraction.asciidoc
+++ b/sources/appendix-value-extraction.asciidoc
@@ -312,10 +312,18 @@ List<String> emails;
 
 When detecting a type argument constraint or cascade, the applicable extractor is determined as follows:
 
-* Choose the most specific extractor matching the type parameter in question;
-an extractor A is more specific than another extractor B if A extracts a subtype of the type extracted by B.
-* If there are several extractors which are equally specific, an `UnexpectedTypeException` is raised.
-TODO: apply rules similar to "ConstraintValidator resolution algorithm".
+* Select all those value extractors which handle a type parameter that maps to the type argument annotated with the constraint or `@Valid` annotation;
+Example:
+** Given `List<@Email String> emails` and considering the default extractors listed above, only the extractors for `List` and `Iterable` are selected.
+The former handles the type parameter `T` of type `List`, which directly maps to the type argument annotated with `@Email`.
+The latter handles the type parameter `E` of type `Iterable` which (indirectly) maps to the annotated type argument (as `List` extends `Iterable` and binds its type parameter `T` to `E` from `Iterable`).
+Other extractors such as the ones for `K` and `V` of `Map` are dismissed, as they handle type parameters not mapping to the annotated type argument
+* From the remaining candidate value extractors chose the one which is most specific to the container type declaring the annoted type argument.
+An extractor A is more specific than another extractor B if A extracts a subtype of the type extracted by B.
+Example:
+** From the two extractors above, the one for `List` is chosen as `List` is a subtype of `Iterable`.
+* If there are several extractors which are equally specific (e.g. several extractors for `List`), an `UnexpectedTypeException` is raised.
+TODO: apply rules similar to "ConstraintValidator resolution algorithm" and further clarify wording
 * When obtaining the extractor for type parameter constraint validation, the declared type of the validated element is used.
 This is to be consistent with constraint validator resolution, which is based on the static type of elements, not the runtime type.
 * When obtaining the extractor for cascaded validation, the runtime type of the cascasded element is used.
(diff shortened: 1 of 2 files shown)


sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,385 @@
+// Bean Validation
+//
+// License: Apache License, Version 2.0
+// See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+
+[[appendix-value-extraction]]
+
+== Value extraction for cascaded validation and type-level constraints (BVAL-508)
+
+This appendix describes the current work-in-progress around the retrieval of values to be validated during cascaded validation and evaluation of type-level constraints (e.g. ). It is based on the original http://beanvalidation.org/proposals/BVAL-508/[proposals for BVAL-508].
```
gsmet: e.g. what?

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,385 @@
...
+
+=== Motivation
+
+Value extraction is needed when constraints are applied to the element(s) stored within a container type. There are two categories:
+
+* Cascaded validation of collections, maps and arrays as triggered via `@Valid`: `@Valid List<Order> orders`; in this case all the values stored in the list must be extracted so each can be validated
+* Validation of type-level constraints as enabled by Java 8: `Optional<@Email String> email` or `Property<@Min(1) Integer> value`; in this case, e.g. the `Integer` wrapped by the property must be extracted so the `@Min` constraint can be applied
```
gsmet: e.g. not necessary here? It seems to be a leftover of a previous formulation.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,385 @@
...
+			i++;
+		}
+	}
+}
+----
+
+[NOTE]
+.Motivation for callback-style API
+====
+Instead of returning the extracted values from the method call, implementations of `ValueExtractor` pass the extracted values to the given receiver callback. This helps to avoid object allocations and allows to handle the case of a single extracted value (`Optional<T`) and multiple extracted values (`List<T>`) in a uniform fashion.
```
gsmet: missing `>`for Optional

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,385 @@
...
+
+* arrays of objects and all primitive data types
+* `java.util.Iterable` (when `@Valid` is given on the iterable element itself, the element and all its entries will be validated; this is to grant backwards compatability with BV 1.1; when `@Valid` is given on the type parameter of an iterable element, all the entries will be validated). A node of type `TYPE_PARAMETER` and with the name "<collection element>" will be added to the path when validating a type parameter constraint.
+* `java.util.Map` (when `@Valid` is given on the map element itself, the element and all its values will be validated; this is to grant backwards compatability with BV 1.1; when `@Valid` is given on the key type parameter, only the keys will be validated; when `@Valid` is given on the value type parameter, only the value will be validated. A node of type `TYPE_PARAMETER` and with the name "<map value>" will be added to the path when validating a type parameter constraint on the map value. A node of type `TYPE_PARAMETER` and with the name "<map key>" will be added to the path when validating a type parameter constraint on the map key.
```
gsmet: s/only the value/only the value**S**/

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,385 @@
...
+* `java.util.Optional`. No node will be appended to the path when validating type-level constraints on `Optional`. Constraints given on an element of type `Optional` apply to the wrapped value by default (see below).
+* `javafx.beans.observable.ObservableValue`. No node will be appended to the path when validating type-level constraints on `ObservableValue`. Constraints given on an element of type `ObservableValue` apply to the wrapped value by default (see below).
+
+==== Plugging in custom extractors
+
+Additional value extractors can be registered with the Bean Validation provider in the following ways:
+
+* Invoke the new method `Configuration#addValueExtractor(ValueExtractor<?>)` (to apply it on the validation factory level)
+* Invoke the new method `ValidatorContext#addValueExtractor(ValueExtractor<?>)` (to apply it for a single `Validator`)
+* Specify the FQN of one or several extractors in `META-INF/validation.xml`:
```
gsmet: Maybe FQCN is better?

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,385 @@
...
+<value-extractors>
+    <value-extractor>com.example.MyExtractor</value-extractor>
+</value-extractors>
+----
+
+==== Retrieval of extractors
+
+When detecting a type-level constraint or cascade, the applicable extractor is determined as follows:
+
+* Chose the most specific extractor matching the type parameter in question; an extractor A is more specific than another extractor B if A extracts a subtype of the type extracted by B.
```
gsmet: s/Chose/Choose/?

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,385 @@
...
+* If there are several extractors which are equally specific, a `UnexpectedTypeException` is raised. TODO: apply rules similar to "ConstraintValidator resolution algorithm".
```
gsmet: a `UnexpectedTypeException` -> a**n** `UnexpectedTypeException`

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,385 @@
...
+----
+
+* The legacy style for cascaded validation is supported as well:
+
+[source,java]
+----
+@Valid List<Order> orders;
+----
+
+This would also validate any constraints on a custom list type (e.g. `MyList#getId()`). TODO: we never clarified that in 1.1. Should it be made explicite?
```
gsmet: s/explicite/explicit/

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,385 @@
...
+
+* Map validation with type-level constraints and cascading:
+
+[source,java]
+----
+@Valid
+Map<@RegExp(...) String, @RetailOrder Order> orders;
+----
+
+This would validate the map's key's against `@RegExp`, the map's values against `@RetailOrder` and apply cascaded validation of the map (values).
```
gsmet: s/key's/keys/

having `(values)` in parenthesis is weird.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,385 @@
...
+
+* When selecting extractors, type parameters must be throroughly traced in the hierarchy. Consider this case where the order of the type parameters of `Map` is swapped in a sub-type:
```
gsmet: you used to write type parameters with an hyphen. I think we should be consistent. Personally, I vote for removing the hyphen. See the glossary here: http://beanvalidation.org/glossary/

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,385 @@
...
+
+Assuming there is no dedicated extractor for `CrazyMap` but only the default ones for `K` and `V` of `Map`, extraction for `@RegExp` must happen via the default map value extractor and extraction for `@Min` via the default map key extractor.
+
+A type-parameter in a sub-type may also map to several type-parameters in a super-type:
+
+[source,java]
+----
+interface NumericMap<T extends Number> extends Map<T, T> {}
+----
+private NumericMap<@Min(1) Integer> integerMap;
```
gsmet: Missing [source,java] formatting.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,385 @@
...
+
+The `@Min` constraint is to be applied to the map' keys and values as the annotated type parameter maps to `K` and `V` of `Map`.
```
gsmet: map' -> map's

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,385 @@
...
+[source,java]
+----
+public class DoubleExtractor implements ValueExtractor<Multimap<@ExtractedValue ?, @ExtractedValue ?>> { ... }
+----
+
+==== XML based configuration
+
+TODO
+
+==== Meta-data retrieval
```
gsmet: Please use metadata in one word.

gsmet:
@gunnarmorling I made a quick proofreading pass. HTH.

sources/appendix-value-extraction.asciidoc
```diff
@@ -386,4 +424,14 @@ TODO
 * `ConstraintsApplyTo` only allows one behavior per annotated element. Should it be per constraint? E.g. for `@NotNull @Email StringProperty email` it may be desirable to apply `@NotNull` to the wrapper but `@Email` to the wrapped value. That's not possible currently.
 * Should type-level constraints be validated when the container is null?
 * What to return from `PropertyDescriptor#getElementClass()` if there is a field of type `Foo` but a getter of type `Optional<Foo>`. So far, BV assumed the types of field and getter to be the same and exposed a single property descriptor (which btw. also may fall apart as of BV 1.1 when the field is of a sub-type of the getter's type). What to return here?
-* Should the presence of type-level constraints alone trigger cascaded validation? It doesn't seem right, but it may be useful when e.g. considering the case of `Tuple` above.
\ No newline at end of file
+* Should the presence of type-level constraints alone trigger cascaded validation? It doesn't seem right, but it may be useful when e.g. considering the case of `Tuple` above.
+* Should we allow extractors to be defined for specific paramterizations of generic types, e.g.
```
gsmet: s/paramterizations/parameterizations/

sources/appendix-value-extraction.asciidoc
```diff
@@ -386,4 +424,14 @@ TODO
...
+
+[source,java]
+----
+public class ListOfIntegerExtractor implements ValueExtractor<List<@ExtractedValue Integer>> { ... }
+
+public class ListOfStringExtractor implements ValueExtractor<List<@ExtractedValue String>> { ... }
+----
+
+I can't see a compelling use case for this (when would extractor behavior differ between different parameterizations of the same generic type) and am leaing towards only supporting the wildcard parameterization (`implements ValueExtractor<List<@ExtractedValue ?>>`).
```
gsmet: s/leaing/leaning/

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+TODO: format of key representations
+* `java.util.Optional`. No node will be appended to the path when validating type-level constraints on `Optional`. Constraints given on an element of type `Optional` apply to the wrapped value by default (see below).
+* `javafx.beans.observable.ObservableValue`. No node will be appended to the path when validating type-level constraints on `ObservableValue`. Constraints given on an element of type `ObservableValue` apply to the wrapped value by default (see below).
+
+==== Plugging in custom extractors
+
+Additional value extractors can be registered with the Bean Validation provider in the following ways:
+
+* Invoke the new method `Configuration#addValueExtractor(ValueExtractor<?>)` (to apply it on the validation factory level)
```
emmanuelbernard: s/on/at/

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+* Invoke the new method `ValidatorContext#addValueExtractor(ValueExtractor<?>)` (to apply it for a single `Validator`)
+* Specify the fully qualified class name of one or several extractors in `META-INF/validation.xml`:
```
emmanuelbernard: Have you considered a service discovery mechanism. I can quickly imagine potential problems with them but that would have been my first thought.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+[source,xml]
+----
+<value-extractors>
+    <value-extractor>com.example.MyExtractor</value-extractor>
+</value-extractors>
+----
+
+==== Retrieval of extractors
+
+When detecting a type-level constraint or cascade, the applicable extractor is determined as follows:
```
emmanuelbernard: It has to be on more than just the presence of the annotation on a type use in particular because of JavaFX types.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+* `java.util.Map` (when `@Valid` is given on the map element itself, the element and all its values will be validated; this is to grant backwards compatability with BV 1.1; when `@Valid` is given on the key type parameter, only the keys will be validated; when `@Valid` is given on the value type parameter, only the values will be validated. A node of type `TYPE_PARAMETER` and with the name "<map value>" will be added to the path when validating a type parameter constraint on the map value. A node of type `TYPE_PARAMETER` and with the name "<map key>" will be added to the path when validating a type parameter constraint on the map key.
+TODO: format of key representations
+* `java.util.Optional`. No node will be appended to the path when validating type-level constraints on `Optional`. Constraints given on an element of type `Optional` apply to the wrapped value by default (see below).
```
emmanuelbernard: Did we discuss the idea of not supporting `@Email Optional<String>` which is I thiink what you describe on 

> Constraints given on an element of type `Optional` apply to the wrapped value by default (see below)

BTW what is the see below that you are referring to? It was not obvious to me.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+<value-extractors>
+    <value-extractor>com.example.MyExtractor</value-extractor>
+</value-extractors>
+----
+
+==== Retrieval of extractors
+
+When detecting a type-level constraint or cascade, the applicable extractor is determined as follows:
+
+* Choose the most specific extractor matching the type parameter in question; an extractor A is more specific than another extractor B if A extracts a subtype of the type extracted by B.
```
emmanuelbernard: Are you talking about the type you extract or the type of the container you extract from?

Are you saying that an extractor of String in Collections is more specific than an extractor of Object in Collections?

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+* If there are several extractors which are equally specific, an `UnexpectedTypeException` is raised. TODO: apply rules similar to "ConstraintValidator resolution algorithm".
+* When obtaining the extractor for type parameter constraint validation, the declared type of the validated element is used. This is to be consistent with constraint validator resolution, which is based on the static type of elements, not the runtime type.
+* When obtaining the extractor for cascaded validation, the runtime type of the cascasded element is used. This is to be consistent with the rules defined for property path construction which are based on the runtime type.
+
```
emmanuelbernard: We need to discuss the wording of these rules, I don't think they are clear.

sources/appendix-value-extraction.asciidoc
emmanuelbernard: I'm not sure I'm reading the rules correctly but can I have one extractor for `Map<key>` and one for `Map<Value>`? It looks like you're saying no.

The corollary is does an extractor extracting from a subtype container override the extractors of the supertype container? e.g. hypothetically extractor on `Set` vs extractor on `Collection`

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+[NOTE]
+.Implementation note
+====
+As extractor retrieval for type parameter constraints is done using the static type of constrained elements, the retrieval can be done once at initialization time and then be cached. This is not possible for retrieval of extractors for cascaded validation.
+====
+
+==== Applying element-level constraints to wrapper elements
+
+Sometimes there is no type parameter to put a constraint to, but still constraints should be applied to the wrapped value instead of the annotated element. JavaFX's property hierachy falls into that category, as it defines specific `Property` sub-types which are not generic:
```
emmanuelbernard: Do we add that people should not use that form when it is possible to annotated the type use location? e.g. `Optional`
Dow e ven forbid it?

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+[source,java]
+----
+List<@NotNull @Email String> emails;
+----
+
+* Cascaded validation of the values in a collection type:
+
+[source,java]
+----
+List<@Valid Order> orders;
```
emmanuelbernard: explain that you will validate the contraints of `Order`

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+* The extractor for cascaded validation is determined based on an element's runtime type:
+
+[source,java]
+----
+Collection<@Valid Order> orders = new ArrayList<>();
+----
+
+Here the most-specific extractor for the runtime type `ArrayList` must be applied, causing the property nodes of violations to have an index set (`Node#getIndex()`).
+
+* The container value passed to a value extractor is retrieved from the element annotated with the constraint or `@Valid`:
```
emmanuelbernard: I don't understand this sentence. you mean the values inside lists are passed to the extractor?

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+----
+
+[source,java]
+----
+public class TupleFirstExtractor implements ValueExtractor<Tuple<@ExtractedValue ?, ?>> {
+
+	@Override
+	public void extractValues(Tuple<?, ?> originalValue, ValueReceiver receiver) {
+		receiver.value( originalValue.getFirst(), "<first>" );
+		}
```
emmanuelbernard: incorrect `}`

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+----
+
+[source,java]
+----
+public class TupleSecondExtractor implements ValueExtractor<Tuple<?, @ExtractedValue ?>> {
+
+	@Override
+	public void extractValues(Tuple<?, ?> originalValue, ValueReceiver receiver) {
+		receiver.value( originalValue.getSecond(), "<second>" );
+		}
```
emmanuelbernard: incorrect `}`

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+----
+
+[source,java]
+----
+private Tuple<@NotNull @Email String, @NotNull @Min(1) Integer> tuple;
+----
+
+==== Invalid examples
+
+* Extractor types must declare exactly one parameter as the extracted value. Hence this extractor type is invalid:
```
emmanuelbernard: is this rule explicit in the formal part of the proposal. I don't think I read it.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+
+TODO
+
+==== Metadata retrieval
+
+TODO
+
+==== Misc.
+
+* Regarding group sequences and default group sequences, the same rules apply for type-level constraints as they apply for regular element-level constraints.
```
emmanuelbernard: Add In particular, you can put the @ConvertGroup contraint to a type-use next to a @Valid. Or something approaching.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+
+=== Open questions
+
+* `ConstraintsApplyTo` only allows one behavior per annotated element. Should it be per constraint? E.g. for `@NotNull @Email StringProperty email` it may be desirable to apply `@NotNull` to the wrapper but `@Email` to the wrapped value. That's not possible currently.
+* Should type-level constraints be validated when the container is null?
```
emmanuelbernard: how would that work? No container, no value to validate against, right?

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+* What to return from `PropertyDescriptor#getElementClass()` if there is a field of type `Foo` but a getter of type `Optional<Foo>`. So far, BV assumed the types of field and getter to be the same and exposed a single property descriptor (which btw. also may fall apart as of BV 1.1 when the field is of a sub-type of the getter's type). What to return here?
+* Should the presence of type-level constraints alone trigger cascaded validation? It doesn't seem right, but it may be useful when e.g. considering the case of `Tuple` above.
```
emmanuelbernard: Why do you think it's useful in the `Tuple` example?

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+* Should we allow extractors to be defined for specific parameterizations of generic types, e.g.
+
+[source,java]
+----
+public class ListOfIntegerExtractor implements ValueExtractor<List<@ExtractedValue Integer>> { ... }
+
+public class ListOfStringExtractor implements ValueExtractor<List<@ExtractedValue String>> { ... }
+----
+
+I can't see a compelling use case for this (when would extractor behavior differ between different parameterizations of the same generic type) and am leaning towards only supporting the wildcard parameterization (`implements ValueExtractor<List<@ExtractedValue ?>>`).
```
emmanuelbernard: I'd say +1 for only supporting the wildcard for now. It looks like the algorithm could be refined if we find a usage later.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+* Choose the most specific extractor matching the type parameter in question; an extractor A is more specific than another extractor B if A extracts a subtype of the type extracted by B.
+* If there are several extractors which are equally specific, an `UnexpectedTypeException` is raised. TODO: apply rules similar to "ConstraintValidator resolution algorithm".
+* When obtaining the extractor for type parameter constraint validation, the declared type of the validated element is used. This is to be consistent with constraint validator resolution, which is based on the static type of elements, not the runtime type.
+* When obtaining the extractor for cascaded validation, the runtime type of the cascasded element is used. This is to be consistent with the rules defined for property path construction which are based on the runtime type.
+
+[NOTE]
+.Implementation note
+====
+As extractor retrieval for type parameter constraints is done using the static type of constrained elements, the retrieval can be done once at initialization time and then be cached. This is not possible for retrieval of extractors for cascaded validation.
+====
```
emmanuelbernard: Add a rule for the following case. When two extractors with the same container type and generic type.

1. probably raise an exception

Do we allow the extraction of non generic type?

```
class WeightedList extends List<String> {
    Integer weight();
}

class WeightExtractor implements ValueExtractor<WeightList<?>> {
    void extractValues(T originalValue, ValueReceiver receiver) {
        receiver.value(originalValue.weight(), "weight");
    }
}
```

If yes, then we might want several of them on the same type.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+		}
+	}
+}
+----
+
+The right callback methods (`indexedValue`, `keyedValue` etc.) must be called in order to allow for proper construction of the property path as per the rules layed out in the BV 1.1 spec.
+
+If a non-null value is passed for `nodeName`, a path node of type `TYPE_PARAMETER` will be appended to the property path. That's desirable for collection types for instance. If null is passed, no node will be appended, resulting in the same path as if the constraint had been given on the element itself instead of a type parameter. That's desirable for pure "wrapper types" such as `Optional`.
+
+In some cases, an extractor applies to the entire extracted type itself (e.g. for arrays). In this case the `@ExtractedValue` annotation is to be given on the extracted type itself:
```
emmanuelbernard: We should capture somewhere (here or in the wiki) why  `@ExtractedValue` lost the ability to refer to a supertype generic type. The main reason being that you can make an extractor for the supertype.

Note that it precludes:

*  a subtype to use better access methods for a faster container navigation (i.e. a better extractor implementation)
* JavaFX  `IntegerProperty` extractor to enforce `@ConstraintsApplyTo(WRAPPED_VALUE)` while JavaFX `Property<?>` extractor to enforce `@ConstraintsApplyTo(ANNOTATED_ELEMENT)`

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+
+As an example, this is how the implementation of the list extractor may look like:
+
+[source,java]
+----
+class ListValueExtractor implements ValueExtractor<List<@ExtractedValue ?>> {
+
+	@Override
+	public void extractValues(List<?> originalValue, ValueReceiver receiver) {
+		int i = 0;
```
gunnarmorling: There was a good reason for it, I'm sure ;)

Let me try @Sanne-style: It'll perform much better if the incoming list is a `LinkedList` ;)

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
+// Bean Validation
+//
+// License: Apache License, Version 2.0
+// See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+
+[[appendix-value-extraction]]
+
+== Value extraction for cascaded validation and type-level constraints (BVAL-508)
```
gunnarmorling: I find these terms more precise as they are using the lingo from the spec.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+
+This appendix describes the current work-in-progress around the retrieval of values to be validated during cascaded validation and evaluation of type-level constraints. It is based on the original http://beanvalidation.org/proposals/BVAL-508/[proposals for BVAL-508].
+
+=== Motivation
+
+Value extraction is needed when constraints are applied to the element(s) stored within a container type. There are two categories:
```
gunnarmorling: I'll update it before pushing.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+
+* Cascaded validation of collections, maps and arrays as triggered via `@Valid`: `@Valid List<Order> orders`; in this case all the values stored in the list must be extracted so each can be validated
```
gunnarmorling: > Also Iterable

Ok.

> Also why isn't Optional<@Valid Foo>

It's considered in the other category.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+* Validation of type-level constraints as enabled by Java 8: `Optional<@Email String> email` or `Property<@Min(1) Integer> value`; in this case, e.g. the wrapped `String` and `Integer` values must be extracted so the `@Email` and `@Min` constraints can be applied
```
gunnarmorling: I mean constraints on a type argument. I'll replace it everywhere.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+
+Both cases can also overlap: `List<@RetailOrder @Valid> retailOrders`.
```
gunnarmorling: Ok.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+
+Type-level constraints and pluggable extractors will also make cascaded validation more flexible. As of BV 1.1, there was a fixed set of data types supported by cascaded validation mandated by the spec (`Iterable`, arrays, `Map` (only the values would be validated)). This excludes use cases such as custom collection types (e.g. Guava's multi-map), validation of map keys or collection types of other JVM languages (such as Ceylon's collection framework).
```
gunnarmorling: Ok.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+
+Allowing to put `@Valid` on type parameters allows to express the subject of cascaded validation more specifically: `Map<@Valid AddressType, @Valid Address> addresses` would describe that both, map keys and values, should be validated.
+
+=== Requirements
+
+1. Constraints can be applied to type parameters (of bean properties, but also method parameters etc.): `Property<@Min(1) Integer> value`
```
gunnarmorling: Nope. Done.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+1. Individual type parameters can be marked for cascaded validation: `Map<@Valid AddressType, Address> addresses`
+1. How values are retrieved must be customizable by means of a value extractor SPI, with a set of defined default extractors
+1. Sometimes constraints should apply to a value wrapped by a container type, but there is no type parameter to put the constraints to. JavaFX's `Property` hierarchy is the most prominent example: `@Email StringProperty emailProperty`. In such case the value should be implicitly extracted from the container if unambiguously doable.
+1. OPEN: Support for validation of nested collections: `List<Map<String, @NotNull String>> addressesByType`.
```
gunnarmorling: > or open for whether or not we should do it?

That. I'm not sure we should. What's your use case for it? It kinda indicates poor modeling to me.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+
+[source,java]
+----
+class CustList extends List<@NotNull Customer> {
+}
+----
+
+=== Solution
+
+Value retrieval for cascaded validation and validation of type-level constraints is done via the `ValueExtractor` SPI:
```
gunnarmorling: Hum, good question. I didn't necessarily meant SPI in terms of the current packaging but rather as "an extension point". I'll change it to API to stay consistent with the current semantics of "SPI" in BV.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+----
+package javax.validation.valueretrieval;
+
+public interface ValueExtractor<T> {
+
+	void extractValues(T originalValue, ValueReceiver receiver);
+
+	interface ValueReceiver {
+
+		void value(Object object, String nodeName);
```
gunnarmorling: No, I haven't explored that any further yet. `Table` is an interesting example, I'll ad that to the todo section.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+			receiver.indexedValue( object, "<collection element>", i );
+			i++;
+		}
+	}
+}
+----
+
+The right callback methods (`indexedValue`, `keyedValue` etc.) must be called in order to allow for proper construction of the property path as per the rules layed out in the BV 1.1 spec.
+
+If a non-null value is passed for `nodeName`, a path node of type `TYPE_PARAMETER` will be appended to the property path. That's desirable for collection types for instance. If null is passed, no node will be appended, resulting in the same path as if the constraint had been given on the element itself instead of a type parameter. That's desirable for pure "wrapper types" such as `Optional`.
```
gunnarmorling: Ok.

sources/appendix-value-extraction.asciidoc
gunnarmorling: > I think eventually @ExtractedValue deserves a dedicated sub-section
Done.

> Any other example besides arrays?

The extractor for plain object associations.

gunnarmorling: Can you do it? I've never seen the motivation behind it.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+.Motivation for callback-style API
+====
+Instead of returning the extracted values from the method call, implementations of `ValueExtractor` pass the extracted values to the given receiver callback. This helps to avoid object allocations and allows to handle the case of a single extracted value (`Optional<T>`) and multiple extracted values (`List<T>`) in a uniform fashion.
+====
+
+==== Default extractors
+
+Compatible implementations provide extractors for the following types out of the box:
+
+* arrays of objects and all primitive data types
```
gunnarmorling: What do you mean exactly? You can have annotations linked to the component type of an array if that's what you are asking:

    String @NotBlank [] names;

Didn't know about that until @gsmet pointed me to it. Thinking of it, the term "type argument constraint" isn't correct in this very case. "type use constraint" probably would, though it's much less digestable (who knows what a "type use" is?).

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+* `java.util.Iterable` (when `@Valid` is given on the iterable element itself, the element and all its entries will be validated; this is to grant backwards compatability with BV 1.1; when `@Valid` is given on the type parameter of an iterable element, all the entries will be validated). A node of type `TYPE_PARAMETER` and with the name "<collection element>" will be added to the path when validating a type parameter constraint.
```
gunnarmorling: > Why not <iterable element>?

Changed. It was `<collection element` in the proprietary implementation in HV.

gunnarmorling: > I'd put the nested points as sublists BTW for clearer reading

Good idea; I've changed this section largely; examples remain to be added.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+* `java.util.Map` (when `@Valid` is given on the map element itself, the element and all its values will be validated; this is to grant backwards compatability with BV 1.1; when `@Valid` is given on the key type parameter, only the keys will be validated; when `@Valid` is given on the value type parameter, only the values will be validated. A node of type `TYPE_PARAMETER` and with the name "<map value>" will be added to the path when validating a type parameter constraint on the map value. A node of type `TYPE_PARAMETER` and with the name "<map key>" will be added to the path when validating a type parameter constraint on the map key.
```
gunnarmorling: > Why validate the Map element itself? In case of sub types of Map with getters?

Yes, right. E.g. `@Min(1) int MyMap#getId()`.

> Is that even the behavior in BV? In HV 5?

It's not clearly stated in BV 1.1. In HV we implemented it that way as that seemed the most reasonable thing to do. I think we got a user report about the lack of it at some point, then we did it.

gunnarmorling: > Why `<map value>` and `<map key>`? what if they are used for other non-map key and value logic?

I'm not following. Can you elaborate?

sources/appendix-value-extraction.asciidoc
gunnarmorling: Yes, we discussed it on and off. I'm open to not supporting it, but then we need it for `@Email StringProperty`. Supporting it in one case but not the other makes some very complex rules, I'm afraid.

> BTW what is the see below that you are referring to?

Section "Applying element-level constraints to wrapper elements". I'll link it.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+TODO: format of key representations
+* `java.util.Optional`. No node will be appended to the path when validating type-level constraints on `Optional`. Constraints given on an element of type `Optional` apply to the wrapped value by default (see below).
+* `javafx.beans.observable.ObservableValue`. No node will be appended to the path when validating type-level constraints on `ObservableValue`. Constraints given on an element of type `ObservableValue` apply to the wrapped value by default (see below).
```
gunnarmorling: A constraint `@Size(min=1) MapProperty myMap` would be applied to the wrapped map. Or do you mean something else?

sources/appendix-value-extraction.asciidoc
gunnarmorling: One reasonable use case is `Optional<List<@Email Strings>> optionalEmails`.

sources/appendix-value-extraction.asciidoc
gunnarmorling: It's not something we are doing in BV so far (apart from provider lookup itself). HV is doing it for constraint validators. You think we should do it? I'll add it to the "open questions" list.

sources/appendix-value-extraction.asciidoc
gunnarmorling: Ok.

gunnarmorling: > I'm not sure I'm reading the rules correctly but can I have one extractor for `Map<key>` and one for `Map<Value>`? It looks like you're saying no.

Yes, you can. An extractor is defined for one specific type parameter.

> The corollary is does an extractor extracting from a subtype container override the extractors of the supertype container?

Yes, it does override. Consider this example:

    List<@Valid Order> orders;

When looking up the extractor, the algorithm will do:

* Narrow the candidate extractors down to those that are declared for type parameter `T` of type `List`; e.g. this would dismiss the extractors for map key and map value but keep the ones for iterable and list.
* Narrow the candidate extractors down to the one(s) for the most specific sub-type; e.g. this would dismiss the one for `Iterable` and keep the one for `List` as the latter is a subtype of `Iterable`
* If there is exactly one extractor remaining, use it. If there are more than one or zero remaining, raise an exception.

sources/appendix-value-extraction.asciidoc
gunnarmorling: I'm talking about the container type targeted by a given extractor. E.g. 

    public class ListValueExtractor implements ValueExtractor<List<@ExtractedValue ?>> { ... }

This extractor is for type `List`.

> Are you saying that an extractor of String in Collections is more specific than an extractor of Object in Collections?

It's an open question, I don't think we should allow extractors for `List<String`, but only for `List<?>`. I.e. wildcard parameterizations only. That's how I've written it up in my local version for now. We can reconsider, but so far I failed to see a use case where it's needed.

sources/appendix-value-extraction.asciidoc
gunnarmorling: Open for suggestions. The example section should clarify it a bit? But yes, the wording isn't 100% spec-proof yet in general. I'm trying to iteratively narrow down the behavior we want without getting too crazy about formal wording for now.

sources/appendix-value-extraction.asciidoc
gunnarmorling: Ok.

sources/appendix-value-extraction.asciidoc
gunnarmorling: > Add a rule for the following case. When two extractors with the same container type and generic type.

Should be covered by "If there are several extractors which are equally specific"?

gunnarmorling: > Do we allow the extraction of non generic type?

Your example isn't fully clear to me. The extractor definition is invalid as `WeightList` has no type parameter. Taking a step back, what's the behavior you'd want to see for `WeightList`?

sources/appendix-value-extraction.asciidoc
gunnarmorling: We can add a recommendation I reckon. I'm not sure whether we can forbid it, as we need to allow it for `StringProperty`, and rules would be complex when allowing it in one case but forbidding it in another.

sources/appendix-value-extraction.asciidoc
gunnarmorling: Ok.

sources/appendix-value-extraction.asciidoc
gunnarmorling: The sentence after the example should clarify it. As `@Valid` for the map values is given on the field, direct field access is used when obtaining the map for passing it to the map value extractor. As `@NotNull` for the map keys is given on the property getter, property access is used when obtaining the map for passing it to the map key extractor.

sources/appendix-value-extraction.asciidoc
gunnarmorling: Ok.

sources/appendix-value-extraction.asciidoc
gunnarmorling: Ok.

sources/appendix-value-extraction.asciidoc
gunnarmorling: Good catch, moved it to the formal part.

sources/appendix-value-extraction.asciidoc
gunnarmorling: Done.

sources/appendix-value-extraction.asciidoc
gunnarmorling: Yes, that looks right.

sources/appendix-value-extraction.asciidoc
gunnarmorling: Cool.

sources/appendix-value-extraction.asciidoc
gunnarmorling: Replaced "type-level" with "type argument" now.

sources/appendix-value-extraction.asciidoc
```diff
@@ -0,0 +1,437 @@
...
+
+public interface ValueExtractor<T> {
+
+	void extractValues(T originalValue, ValueReceiver receiver);
+
+	interface ValueReceiver {
+
+		void value(Object object, String nodeName);
+
+		void iterableValue(Object object, String nodeName);
```
gunnarmorling: Can you elaborate on "key object to path node converter"?

sources/appendix-value-extraction.asciidoc
gunnarmorling: Added to the open questions.

gunnarmorling:
@emmanuelbernard Thanks for reviewing. I've just pushed an update which should either address your remarks or capture them at the end of the document.

Most notably, I've expended the sections on default extractor behavior and extractor resolution. I think with that it's good to go for the appendix and we can refine from there and answer the open questions. Can you take another look and let me know of your thoughts?

sources/appendix-value-extraction.asciidoc
emmanuelbernard: @gunnarmorling I was thinking of offering a contract to convert an object into a `Path.node`

```
class  GuavaTableValueExtractor implements ValueExtractor {
    void extractValues(GuavaTable originalValue, ValueReceiver receiver);
        Set< Table.Cell<R, C, V> > cells = originalValue.cellSet()
        for (Table.Cell cell: cells) {
            receiver.valueWithKey( cell.getValue(), cell );
        }

    }
}

class CellToKeyPathNodeConverter implements PathNodeConverter<Table.Cell> {
    NodeKey convert(TableCell cell) {
        return new NodeKey() {
              String getName() {
                  return "cell.getRow() + ", " + cell.getColumn();
                   // TODO what is the prefix, ideally I'd like a path like tables[row1,column1]
              }
    }
}
```

But thinking about it more:

* it ends up being just a string reprensentation so maybe the string param is enough
* sharing implementations of object to `Path.Node` ends up ambiguous (several version of the same object type might have different representations) so putting the impl in the `ValueExtractor` is probably better

But I'm still confused as to what the nodes would look like both in string and in actual `Path.Node` approach. Examples would help.

sources/appendix-value-extraction.asciidoc
emmanuelbernard: well but what if @Valid is on the type use. That's what makes the sentence confusing I guess. For sure I was light years away from your explanation when I read it.

sources/appendix-value-extraction.asciidoc
emmanuelbernard: I thought somehow that there would be a need for a type parameter node in our tree but this is not necessary (at least for the error report), as the node points to the value and not the annotation holder itself. I got confused.

Maybe metadata will be a different story.

sources/appendix-value-extraction.asciidoc
emmanuelbernard: My comment was more related to the use of of the `ValueReceiver` method like `keyValue` `indexValue` which would lead to it displaying a `<map key`> node name even if that's not a map that calls it

sources/appendix-value-extraction.asciidoc
emmanuelbernard: @gunnarmorling I don't think I agree with your line of reasoning.
You're treating Iterable and sons + Map as a special case where `@Email Map<String, String> emails` would not be possible.
But you're treating `Optional` differently.
It's not clear why following what I think is your reasoning

sources/appendix-value-extraction.asciidoc
emmanuelbernard: I think we should do it at first view. But the problem arises if there is a need to disable or override some extractors with others. Which might make it a nogo.

Still, having to enable my guava extractors manually feels annoying. You could argue that HV would have the Guava extractors enable ootb but then you're abck to having to find a solution to the problem I describe above. At this stage we could solve it in a generic way.

sources/appendix-value-extraction.asciidoc
emmanuelbernard: > Yes, you can. An extractor is defined for one specific type parameter.
> [...]
> Yes, it does override. Consider this example:

note that this was not my interpretation of the reading, better making it more explicit.

sources/appendix-value-extraction.asciidoc
emmanuelbernard: The reasons I ask is that the sentence made me think you were mentioning the the type you extract, not the container type. because you mentioned "the type parameter in question"

sources/appendix-value-extraction.asciidoc
emmanuelbernard: No mental energy to do it now, add a TODO to remember.

sources/appendix-value-extraction.asciidoc
emmanuelbernard: I think you answered the question by no. You do not want the mechanism to extract values that are not related to a type parameter.

I'm only slightly uneased by it as it might block us in a future where we want to use the extractor for more things that type parameter validation and `@Valid`. I don't have concrete examples of such future however.

sources/appendix-value-extraction.asciidoc
emmanuelbernard: Would they be that complex? If the type param is expressed in the declaration, use it or else wait for the exception police :)

gunnarmorling: Ok, I've added it to the questions. It seems a somewhat arbitrary restriction I think.

sources/appendix-value-extraction.asciidoc
gunnarmorling: Note you can declare `class WeightExtractor implements ValueExtractor<@ExtractedValue WeightList> { ... }`.

sources/appendix-value-extraction.asciidoc
gunnarmorling: I've extended that a bit now.

sources/appendix-value-extraction.asciidoc
gunnarmorling: Ok, I've expanded it further.

sources/appendix-value-extraction.asciidoc
emmanuelbernard: @gunnarmorling the spec does not explain what that does though, right?

sources/appendix-value-extraction.asciidoc
gunnarmorling: I've added it as a question whether to support `@Email Optional<String>` or not.

> You're treating Iterable and sons + Map as a special case where @Email Map<String, String> emails would not be possible.

Hum, no, nothing special really. Check out the expanded section on "Retrieval of extractors" ("When detecting a regular element-level constraint (i.e. non type argument constraint)...").  `@Email Map<String, String>` isn't possible because it's not clear to why type parameter (key or value) the constraint should be applied. Hence giving the constraint on the outside is only supported if there is exactly one type parameter in the hierarchy (`T` of `List` mapping to `E` of `Iterable`  counts as one).

sources/appendix-value-extraction.asciidoc
gunnarmorling: Hum, not sure whether I get behind that concern. Of course each extractor must make sure it only calls the right receiver methods. E.g. for legacy style `@Valid Map` it'd call once `value()` (for the map object itself) and then for each entry `keyedValue()`.

sources/appendix-value-extraction.asciidoc
gunnarmorling: > I thought somehow that there would be a need for a type parameter node in our tree

There is. We'd have a node of kind `TYPE_ARGUMENT` and with the index of the concerned array element.

sources/appendix-value-extraction.asciidoc
gunnarmorling: We say

> In some cases, an extractor applies to the entire extracted type itself (e.g. for arrays). In this case the @ExtractedValue annotation is to be given on the extracted type itself

emmanuelbernard: ok so WeightExtractor in this case would extract weight and no longer map keys and map values.

sources/appendix-value-extraction.asciidoc
emmanuelbernard: `ValueReceiver.value()` does not allow you to pass a "key" (e.g. Guava Table row+column).
such a map node as a `getKey()` method that we would not have a cousin for.

sources/appendix-value-extraction.asciidoc
emmanuelbernard: Ah I got it now. Then it should be named `TYPE_ARGUMENT`, see http://beanvalidation.org/glossary/

emmanuelbernard: BTW how do you answer this message thread on "outdated" references? The email itself gives no context and then I ahve to manuyally open all the outdated divs to get the context and reply to you. Soooo annoying.

sources/appendix-value-extraction.asciidoc
gunnarmorling: There is `keyedValue()`, but to begin with, I'm not sure how the resulting node should look like for `Table`. Esp. we don't have a way to create a combined representation of row and column as they'd be handled by two different extractors.

The current receiver methods just expose what's available through `Node` currently. Let's discuss this case in tomorrow's meeting?

sources/appendix-value-extraction.asciidoc
gunnarmorling: Yes, I've already renamed it into `TYPE_ARGUMENT`.

On the outdated references, it sucks big time. Don't have a better solution than what you do, apart from continuing the discussion elsewhere?

gunnarmorling: Btw. "type argument" doesn't fully cut it for the array case (as I stated somewhere before). There is no such thing in this case. "type use" would be the more general and correct term IIUC.

sources/appendix-value-extraction.asciidoc
emmanuelbernard: ah right, I was too focused on validating what's inside the map (i.e. two levels down) and forgot that case.

sources/appendix-value-extraction.asciidoc
emmanuelbernard: I'm not concerned too much about the node generated by the row extractor or the column extractor, I imagine these will do something like `valuereceiver.value(row, "<table row>")`

 I'm concerned about the `Table` value extractor because in essence you want the row and column to be part of the path to identify the value. Just like our node for Map value has `getKey()`. While pondering the problem I saw a few initial approaches:
- make nodes generified and offer a way to get to a generic `getKey` (in the general sense of key).
- keep `getKey` as is and return `Object`

In both case, the key object would need to be a composite key if you will

```
class RowAndColumnKey {
    Object getRow();
    Object getColumn();
    String toString(); //cool for the string reprensetation
}

gunnarmorling:
Rebased and merged. Thanks a ton for doing the review dance, @emmanuelbernard!

PR #120, opened 2017-01-05
BVAL-508 Use size() instead of length for the list example

gsmet:


sources/appendix-value-extraction.asciidoc
```diff
@@ -86,7 +86,7 @@ class ListValueExtractor implements ValueExtractor<List<@ExtractedValue ?>> {
 
 	@Override
 	public void extractValues(List<?> originalValue, ValueReceiver receiver) {
-		for ( int i = 0; i < originalValue.length; i++ ) {
+		for ( int i = 0; i < originalValue.size(); i++ ) {
```
gunnarmorling: That example was using Groovy and `length` has been added dynamically ;)

PR #123, opened 2017-01-10
BVAL-554 Add the ability to generate a single asciidoc file containing all the specification

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-554

I added a pom file so that it's possible to quickly import the asciidoctor extensions in Eclipse.


gunnarmorling:
Nice, thanks! I'll play around a bit with it tomorrow.

gsmet:
Looks like this one can be closed.

PR #126, opened 2017-01-16
BVAL-508 Misc. clarifications:

gunnarmorling:
* Usage of @ExtractedValue for array value extractors
* Exception model


gsmet:
Added one more commit and merged! Thanks!

PR #128, opened 2017-01-19
BVAL-513 Add the Google Analytics marker to the spec

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-513

For an unknown reason, I didn't get to put the marker id in the configuration. Looks like there's something weird going on.

For now, this will do.

It looks like it's an issue to put the docinfo files outside the asciidoc source directory (I have the same issue with the HV documentation). Moreover, there is an additional issue here as putting the docinfo file in the source directory does not work at all.


gunnarmorling:
Rebased and applied. Thanks, @gsmet!

PR #129, opened 2017-01-20
BVAL-508 Updating value extraction proposal as per latest discussions;

gunnarmorling:
* Using payloads instead of @ConstraintsApplyTo annotation
* Dedicated @UnwrapByDefault annotation
* Documenting agreement on some of the open questions

### DIFF
--- a/sources/appendix-value-extraction.asciidoc
+++ b/sources/appendix-value-extraction.asciidoc
@@ -8,7 +8,8 @@
 [appendix]
 == Value extraction for cascaded validation and type argument constraints (BVAL-508)
 
-This appendix describes the current work-in-progress around the retrieval of values to be validated during cascaded validation and evaluation of type argument constraints.
+This appendix describes the current work-in-progress around the retrieval of values to be validated during cascaded validation (e.g. `@Valid List<Order> orders`)
+and evaluation of type argument constraints (e.g. `Optional<@Email String> email`)
 It is based on the original http://beanvalidation.org/proposals/BVAL-508/[proposals for BVAL-508].
 
 === Motivation
@@ -77,7 +78,7 @@ public interface ValueExtractor<T> {
 }
 ----
 
-An extractor is tied to one specific type parameter of the type from which it extracts values. The `@ExtractedValue` annotation is used to mark that type parameter. 
+An extractor is tied to one specific type parameter of the type from which it extracts values. The `@ExtractedValue` annotation is used to mark that type parameter.
 
 As an example, this is how the implementation of the list extractor may look like:
 
@@ -259,7 +260,7 @@ Additional value extractors can be registered with the Bean Validation provider 
 
 * Invoke the new method `Configuration#addValueExtractor(ValueExtractor<?>)` (to apply it at the validation factory level)
 * Invoke the new method `ValidatorContext#addValueExtractor(ValueExtractor<?>)` (to apply it for a single `Validator`)
-* Specify the fully qualified class name of one or several extractors in `META-INF/validation.xml`: 
+* Specify the fully qualified class name of one or several extractors in `META-INF/validation.xml`:
 
 [source,xml]
 ----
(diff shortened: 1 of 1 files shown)


sources/appendix-value-extraction.asciidoc
```diff
@@ -280,30 +281,38 @@ JavaFX's property hierarchy falls into that category, as it defines specific `Pr
...
 
+====
+
 [source,java]
 ----
-public enum ValidationTarget {
-    ANNOTATED_ELEMENT,
-    WRAPPED_VALUE;
-}
+@Email(payload=Unwrapping.Unwrap.class) StringProperty emailProperty;
```
gsmet: Very minor gripe, but some padding around the = would be nice to be consistent with the rest.

This is a bad example as `StringProperty` will use the automatic unwrapping. Maybe, simply call it `StringWrapper` (it is used in an example below) or something similar?

sources/appendix-value-extraction.asciidoc
```diff
@@ -316,18 +325,17 @@ class PropertyExtractor implements ValueExtractor<Property<@ExtractedValue ?>> {
...
 
 In rare cases it may happen that a constraint should be applied to the wrapped value although an extractor exists.
-In this case the `ConstraintsApplyTo` annotation can be given on the element itself to prevent the unwrapping:
+In this case the `Unwrapping.Skip` payload type can be specified for that constraint to prevent the unwrapping:
 
 [source,java]
 ----
-@ConstraintsApplyTo(ANNOTATED_ELEMENT)
-@NotNull StringProperty emailProperty;
+@NotNull(payload=Unwrapping.Skip.class) StringProperty emailProperty;
```
gsmet: Same remark about the padding around =

sources/appendix-value-extraction.asciidoc
```diff
@@ -339,8 +347,7 @@ over
 
 [source,java]
 ----
-@Email
-@ConstraintsApplyTo(WRAPPED_VALUE)
+@Email(payload=Unwrapping.Unwrap.class)
```
gsmet: Ditto.

sources/appendix-value-extraction.asciidoc
```diff
@@ -393,14 +400,14 @@ When detecting a regular element-level constraint (i.e. non type argument constr
...
+1. If the constraint carries the `Unwrapping.Unwrap` payload:
 ** If no type parameter or more than one type parameter was found in step 2, raise a `ConstraintDeclarationException`
 ** Choose the most specific extractor matching the single type parameter found in step 2
 ** If there are several extractors which are equally specific, a `ConstraintDeclarationException` is raised.
 
-1. If `@ConstraintsApplyTo` is not given:
+1. If the constraint neither carries the `Unwrapping.Unwrap` nor the `Unwrapping.Skip` payload:
 ** If no type parameter or more than one type parameter was found in step 2, don't apply any value extractor
-** Choose the most specific extractor matching the single type parameter found in step 2 which is marked with `@ConstraintsApplyTo(WRAPPED_VALUE)`
+** Choose the most specific extractor matching the single type parameter found in step 2 which is marked with `UnwrapByDefault`
```
gsmet: Is it really what we want to do? I mean I would expect us to choose the most specific extractor AND unwrap only if it is marked with `UnwrapByDefault`.

sources/appendix-value-extraction.asciidoc
```diff
@@ -650,21 +656,20 @@ private CachableValue<@Min(1) Integer> foo;
...
-* Element-level constraints cannot be applied if there is no type parameter at all or multiple non-mapping type parameters in the annotated element's type hierarchy. Thus an exception will be raised in the following cases:
+* Element-level constraints cannot be applied if there is no type parameter at all or multiple non-mapping type parameters in the annotated element's type hierarchy.
+Thus an exception will be raised in the following cases:
 +
 [source,java]
 ----
 // no type parameter
-@ConstraintsApplyTo(WRAPPED_VALUE)
-@Email
+@Email(payload=Unwrapping.Unwrap.class)
```
gsmet: Padding.

sources/appendix-value-extraction.asciidoc
```diff
@@ -650,21 +656,20 @@ private CachableValue<@Min(1) Integer> foo;
...
 private String email;
 ----
 +
 [source,java]
 ----
 // multiple type parameters
-@ConstraintsApplyTo(WRAPPED_VALUE)
-@Email
+@Email(payload=Unwrapping.Unwrap.class)
```
gsmet: Padding.

sources/appendix-value-extraction.asciidoc
```diff
@@ -728,25 +740,27 @@ So far, BV assumed the types of field and getter to be the same and exposed a si
...
++
+__We agreed to require `@Valid` for the sake of consistency.__
 
-1. For an element with a type argument, should it be allowed to specify constraints on the element (and use `@ConstraintsApplyTo(WRAPPED_ELEMENT)`) or should it be disallowed?
+1. For an element with a type argument, should it be allowed to specify constraints on the element (and use the `Unwrapping.Unwrap` payload) or should it be disallowed?
 +
 [source,java]
 ----
-@Email @ConstraintsApplyTo(WRAPPED_ELEMENT) Optional<String> email;
+@Email(payload=Unwrapping.Unwrap.class) Optional<String> email;
```
gsmet: Padding.

sources/appendix-value-extraction.asciidoc
gunnarmorling: Ok.

sources/appendix-value-extraction.asciidoc
gunnarmorling: Agreed.

sources/appendix-value-extraction.asciidoc
gunnarmorling: About `StringProperty`, I see where you are coming from, but I'll leave it as is for now. Automatic unwrapping is only introduced in detail later on, so I think it's alright. We can think of another example when putting it into the individual spec chapters.

gsmet:
Fixed a typo introduced in the latest changes, squashed and merged. Thanks!

sources/appendix-value-extraction.asciidoc
gsmet: Ah sorry, just saw your comment now (it was hidden by GitHub) :/. I thought you forgot about it so I changed it before merging. I used StringWrapper as in another example. Feel free to change it again if you feel strongly about it.

PR #134, opened 2017-02-14
BVAL-561 Make @OverridesAttribute repeatable and documented

kasabuta4:
https://hibernate.atlassian.net/projects/BVAL/issues/BVAL-561

PR #135, opened 2017-02-15
BVAL-561

gunnarmorling:
Supersedes #134.

### DIFF
--- a/sources/constraint-definition.asciidoc
+++ b/sources/constraint-definition.asciidoc
@@ -409,7 +409,7 @@ include::{spec-examples-source-dir}org/beanvalidation/specexamples/constraintdef
 
 ====
 
-[tck-testable]#It is possible to override attributes and messages defined on a composing annotation. An attribute from the main annotation is used to override one or more attributes of the composing annotations. Such an attribute is annotated with the [classname]`@OverridesAttribute` annotation or its multivalued equivalent [classname]`@OverridesAttribute.List`.#
+[tck-testable]#It is possible to override attributes and messages defined on a composing annotation. An attribute from the main annotation is used to override one or more attributes of the composing annotations. Such an attribute is annotated with the [classname]`@OverridesAttribute` annotation or its multivalued equivalent [classname]`@OverridesAttribute.List`.([classname]`@OverridesAttribute` is meta-annotated with [classname]`java.lang.annotation.Repeatable`, referencing the [classname]`@OverridesAttribute.List`, so, simply repeating the [classname]`@OverridesAttribute` annotation is possible.)#
 
 [[example-composing-overridden]]
 
--- a/sources/constraint-definition.asciidoc
+++ b/sources/constraint-definition.asciidoc
@@ -409,7 +409,12 @@ include::{spec-examples-source-dir}org/beanvalidation/specexamples/constraintdef
 
 ====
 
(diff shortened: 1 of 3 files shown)


sources/constraint-definition.asciidoc
```diff
@@ -455,7 +460,18 @@ if [classname]`@FrenchZipCode` is defined as
...
+[tck-testable]#If a constraint is used more than once as a composing constraint, the multi value constraints model as described in <<constraintsdefinitionimplementation-multipleconstraints>> is used.#
+
+To select a specific composing constraint, [methodname]`OverridesAttribute.constraintIndex` is used.
+[tck-testable]#If the composing constraints are directly given on the composing constraint, `constraintIndex` refers to the left-to-right order of the constraints of this type in which they are given on the composing constraint.#
+[tck-testable]#If the composing constraints are specified using their corresponding `List` annotation, `constraintIndex` refers to the index within the `value` array.#
+[tck-testable]
+--
+A composing constraint must not be given directly on the composing constraint and using the corresponding `List` annotation at the same time.
```
gsmet: Can we determine that?

kasabuta4: According to Java Language Specification 8 Section 9.7.5, applying repeatable annotations and their container annotation at the same time is compile-time error.
So, we can determine that.

https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.7.5

gunnarmorling: Actually it is allowed to have one instance of the repeatable annotation type and one instance of the container annotation. See the second example in the section you linked:

> This rule is designed to allow the following code:
>
> @Foo(1) @FooContainer({@Foo(2)})
> class A {}

But we can determine that as `AnnotatedElement#getDeclaredAnnotations()` returns the _directly present_ annotations, e.g. `@Pattern` and `@Pattern.List`. In this case an exception would have to be thrown.

spec-examples/src/test/java/org/beanvalidation/specexamples/constraintdefinition/constraintcomposition/overriding/constraintindex/EmmanuelsEmail.java
```diff
@@ -34,12 +34,10 @@
...
 //tag::include[]
-@Pattern.List({
-		@Pattern(regexp = "[A-Z0-9._%+-]+@[A-Z0-9.-]+\\.[A-Z]{2,4}"), // email
-		@Pattern(regexp = ".*?emmanuel.*?") // emmanuel
-})
-@Constraint(validatedBy = {})
 @Documented
+@Constraint(validatedBy = {})
+@Pattern(regexp = "[A-Z0-9._%+-]+@[A-Z0-9.-]+\\.[A-Z]{2,4}"), // email
+@Pattern(regexp = ".*?emmanuel.*?") // emmanuel
```
kasabuta4: remove a comma(,) between `@Pattern` annotations.

kasabuta4:
Thanks for your revise.
I have checked it out, and it seems all right to me,
except for removing a comma(,) between `@Pattern` annotation in EmmanuelsEmail.java.

gunnarmorling:
Thanks, good catch! Seems I didn't compile the example :)

gunnarmorling:
Fixed the example. Thanks again, @kasabuta4!

PR #137, opened 2017-02-18
BVAL-568 Make built-in annotations documented

kasabuta4:
https://hibernate.atlassian.net/browse/BVAL-568


gunnarmorling:
Merged; thanks!

kasabuta4:
appendix-value-extraction.asciidoc copies, not refer to,  @ExtractedValue source code, so we have to change it, too.
But, I suppose appendix-value-extraction.asciidoc will be revised soon, so leave it as it is.

Thanks!

gunnarmorling:
Right, the appendix will go away soon and the contents will be merged into
the right parts of the spec. Then we can change it to an include, too.

PR #140, opened 2017-02-20
BVAL-571 Adding service loader mechanism for providing value extractors

gunnarmorling:


sources/appendix-value-extraction.asciidoc
```diff
@@ -256,18 +256,26 @@ TODO
...
 
-Additional value extractors can be registered with the Bean Validation provider in the following ways:
-
-* Invoke the new method `Configuration#addValueExtractor(ValueExtractor<?>)` (to apply it at the validation factory level)
-* Invoke the new method `ValidatorContext#addValueExtractor(ValueExtractor<?>)` (to apply it for a single `Validator`)
-* Specify the fully qualified class name of one or several extractors in `META-INF/validation.xml`:
+Additional value extractors can be plugged in when bootstrapping a `Validator` or `ValidatorFactory`, amending and/or overriding the set of built-in value extractors.
+The following ways to do so exist, in descending order of precedence:
 
+* Invoking the new method `ValidatorContext#addValueExtractor(ValueExtractor<?>)` (to apply it for a single `Validator`)
```
gsmet: Really nitpicking but I don't think we should use `new` in the spec. I would simply remove the word here and in the line below.

gunnarmorling:
Agreed. The wording in the appendix still is less formalized than it will
be when it's in the actual spec sections. Latest I'd have removed the "new"
when doing that move.

sources/appendix-value-extraction.asciidoc
```diff
@@ -256,18 +256,26 @@ TODO
...
 <value-extractors>
     <value-extractor>com.example.MyExtractor</value-extractor>
 </value-extractors>
 ----
+* Placing a file named _META-INF/services/javax.validation.valueextraction.ValueExtractor_ on the classpath,
+with the fully-qualified name(s) of one or more extractor implementations as its contents (i.e. employing the Java service loader mechanism)
+
+A value extractor for a given type and extracted type parameter specified at a higher level overrides any other extractors for the same type and extracted type given at lower levels.
+If e.g. a `ValueExtractor<List<@ExtractedValue ?>>` is given via `ValidatorContext#addValueExtractor(ValueExtractor<?>)`,
+it will take precedence of any value extractor for the same type (`List`) and type parameter (`E`) given via `Configuration#addValueExtractor(ValueExtractor<?>)`,
```
emmanuelbernard: I'm a bit confused why you mention `E` whereas the definition mentions `?` right before. So it's unclear what behavior is expected in the end.

gunnarmorling: The extractor definition would be `MyListExtractor implements ValueExtractor<List<@ExtractedValue ?>> { ... }`. Then `@ExtractedValue` is applied to the type parameter `E` of the `List` interface. I.e. `?` represents `E`.

emmanuelbernard: Hopefully, we can find a clearer wording :)

gsmet:
@gunnarmorling what's the status of this? Do you plan to make changes now following Emmanuel's comments or do you want to do it later when injecting the append in the spec?

gunnarmorling:
I'll update the wording and then push it.

PR #143, opened 2017-02-27
BVAL-560 Mention Oracle instead of Sun in the evaluation license

gsmet:
@gunnarmorling I also switched back to the Apache License. As I understood it, the snapshots are covered by the AL and the official releases by the evaluation license?

Just cherry-pick the first commit if I'm wrong.

(Btw, I added `and/or its affiliates` to conform with the Oracle trademarks guidelines)


gunnarmorling:
Thanks, rebased and applied. I didn't apply the license change, let's decide on that one once we do another release (either with or without "official" ED).

PR #144, opened 2017-03-03
BVAL-577 Proof-read "3.3. Constraint composition" on "composing" vs. "composed"

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-577

### DIFF
--- a/sources/constraint-definition.asciidoc
+++ b/sources/constraint-definition.asciidoc
@@ -436,7 +436,7 @@ If left undefined, the default value for [methodname]`@OverridesAttribute.name` 
 
 [NOTE]
 ====
-[tck-testable]#A composing constraint can itself be a composed constraint. In this case, attribute values are overridden recursively according to the described rules.# Note however, that a forwarding rule (as defined by [classname]`@OverridesAttribute`) is only applied to the direct composing constraints.
+[tck-testable]#A composed constraint can itself be a composing constraint. In this case, attribute values are overridden recursively according to the described rules.# Note however, that a forwarding rule (as defined by [classname]`@OverridesAttribute`) is only applied to the direct composing constraints.
 ====
 
 Using <<example-composing-overridden>>,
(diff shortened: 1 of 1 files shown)


sources/constraint-definition.asciidoc
```diff
@@ -436,7 +436,7 @@ If left undefined, the default value for [methodname]`@OverridesAttribute.name`
 
 [NOTE]
 ====
-[tck-testable]#A composing constraint can itself be a composed constraint. In this case, attribute values are overridden recursively according to the described rules.# Note however, that a forwarding rule (as defined by [classname]`@OverridesAttribute`) is only applied to the direct composing constraints.
```
gunnarmorling: Wasn't this correct? A constraint given as meta-annotation (=composing constraint) can be composed from others itself.

gsmet: I understood it the other way around: that a meta annotation can also be used to compose other meta annotations. I think it's more logical to state it this way.

gunnarmorling: Let's discuss an example to make sure we are on the same page on the terms:

    @NotNull
    @Size(min=1)
    public @interface NotEmpty {
        ...
    }

Here I'd consider `@NotEmpty` as the composed constraint and `NotNull`/`@Size` as the composing constraints. Of course `@NotNull` could be used to compose other constraints than `@NotEmpty` (which seems to be what you are saying), but that's not what this section is about IIUC. Instead it's there to say that `@NotNull` and `@Size` could be composed constraints themselves. In that light, the original version was correct. Does that make sense?

gsmet: Well, I think it makes sense either way. What I'm saying is that `@NotEmpty` can also be used as a composing constraint. It's usually the way you would have it: you have an existing composed constraint and you want to use it as a composing constraint.

I think it makes more sense this way but I don't have a strong opinion about it. Feel free to ignore this part if you prefer the existing version. Both are correct.

gunnarmorling: Ah, got you. I think it becomes more clear that it was meant as I understood it (and as it was written originally) when looking at the next sentences which describe the attribute overriding. I'll exclude that specific change and merge the PR.

gunnarmorling:
Applied with the small rework discussed above. Thanks, @gsmet!

PR #145, opened 2017-03-07
BVAL-578 Make the assertion related to NoProviderFoundException not testable

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-578


gsmet:
@gunnarmorling I gave it some thoughts yesterday evening and, in the end, I think we should even remove it from the TCK assertions altogether.

WDYT?

gunnarmorling:
I would leave the assertion, but mark it as not testable, as you did.

PR #147, opened 2017-03-20
BVAL-580 Prepare the migration to string ids spec-wise

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-580


gunnarmorling:
@gsmet Could you rebase this one?

tck-sectionid-renaming.xsl
```diff
@@ -0,0 +1,117 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ JBoss, Home of Professional Open Source
+  ~ Copyright 2012, Red Hat, Inc. and/or its affiliates, and individual contributors
```
gunnarmorling: The short header could be used.

gsmet:
@gunnarmorling done

tck-sectionid-renaming.xsl
```diff
@@ -0,0 +1,117 @@
...
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<!--
+  ~ Creates the sed command file for renaming the section ids to their string values
+  ~
+  ~ @author Guillaume Smet
+  -->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xalan="http://xml.apache.org/xslt"
```
gunnarmorling: Do we need this only once? Or repeatedly? I'm not fully clear about the "TCK workflow", but I trust you are on top of it.

tck-sectionid-renaming.xsl
gsmet: Yeah, I plan to do a clean up in the spec regarding that. I chose to be consistent and change everything at the same time.

tck-sectionid-renaming.xsl
gsmet: We only need it once. I was a bit worried about not committing it somewhere in case we would have to revisit this subject.

tck-sectionid-renaming.xsl
gsmet: Created https://hibernate.atlassian.net/browse/BVAL-584 so I don't forget about it.

tck-sectionid-renaming.xsl
gunnarmorling: Ok; going to merge it.

gunnarmorling:
@gsmet Looking at the generated audit file, it's containing some new empty sections now (e.g. for examples). I think you mentioned it, but I don't quite remember the details. Is it we cannot suppress them? Or they are needed somehow?

PR #148, opened 2017-03-21
BVAL-583 Use the common theme for the asciidoc output

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-583

I had prepared this for a while. FWIW, I didn't include the logo in the HTML as I did it for the TCK documentation so that the spec is still self contained.


gunnarmorling:
> so that the spec is still self contained

WDYM by that?

gunnarmorling:
Rebased and applied. Thanks, @gsmet!

gsmet:
So that we don't have to have the logo on the side. I think it's better if the spec is all self contained and you don't need additional files (logo, CSS, whatever).

gunnarmorling:
Ah, you mean the rendered output. Ok.

PR #150, opened 2017-03-27
BVAL-528 Replace API listings by source includes

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-528

FYI, there are a couple of things I didn't replace: sometimes, we extract part of the API in a listing and I wasn't too comfortable with this. We might revisit it later though. There is a good example of this here: http://beanvalidation.org/latest-draft/spec/#boostrapping-validation .


gunnarmorling:
Merged; thanks, @gsmet!

PR #153, opened 2017-04-19
BVAL-591 Adding container element validation to actual spec chapters, dissolving appendix

gunnarmorling:
~~This still needs some more examples and the pending metadata API extension, but it should be good for review otherwise.~~

Feedback welcome!


sources/xml-descriptor.asciidoc
```diff
@@ -498,6 +500,18 @@ Other custom elements of an annotation are represented by `element`. [tck-testab
...
+        <!-- Map<@NotBlank String, @Valid PhoneNumber>
+         -->
+        <field name="phoneNumbersByType">
+            <container-element-type type-argument-index="0">
+                <constraint annotation="javax.validation.constraints.NotBlank"/>
+            </container-element-type>
+            <container-element-type type-argument-index="1">
+                <valid/>
+            </container-element-type>
+            <valid/>
```
gsmet: Is the last `<valid/>` really wanted?

gunnarmorling: Nope, good catch!

gsmet:
Rebased and merged, nice work!

sources/constraint-declaration-validation.asciidoc
```diff
@@ -11,6 +11,9 @@ The Bean Validation specification defines a framework for declaring constraints
 
 Bean Validation also offers a way to declare constructor and method constraints where parameters and return values are the constrained elements. We will discuss method constraints declaration in detail in <<constraintdeclarationvalidationprocess-methodlevelconstraints>>.
 
+Furthermore, constraints can be applied to the elements of generic container types such as `Map`, `List` or `Optional`.
```
emmanuelbernard: I'm tempted to say that's it's closer to a "normal" declaration and this should be expressed before method level constraints.

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
 
 The return value constraints in [classname]`DefaultOrderService` are legal, as they strengthen the postconditions of the [methodname]`placeOrder()` method as constituted by the super class [classname]`OrderService` but don't weaken them.
 
+[[constraintdeclarationvalidationprocess-containerelementconstraints]]
+=== Container element constraints
+
+Constraints can be applied to the elements of generic containers, e.g. `List`, `Map` or `Optional`.
+This is done by putting constraint annotations to the type arguments of such containers.
+Container element constraints can be applied to containers that are
```
emmanuelbernard: I find this sentence a tiny bit ambiguous. I read it as the CE constraints apply to the container which is not what you want to say.

Maybe

"Containers considered can be in the following locations:"

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+    [...]
+}
+
+public List<@NotBlank String> getMatchingRecords(List<@NotNull @Size(max=20) String> searchTerms) {
+    [...]
+}
+----
+====
+
+When a field, property, executable parameter or method return value is validated,
```
emmanuelbernard: "and that this value if of a container type,"

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+any container element constraints of that element will be validated alongside any other constraints hosted by that element.
```
emmanuelbernard: This sentence is losing me. Are you trying to say

> When a field, property, executable parameter or method return value is validated and that this value is of a container type, then all values contained in the container will be validated provided that their container element is constrained

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+For container element constraints, the same rules for validation groups and group sequences apply as for any other constraint on the same element.
+
+When a container element constraint is detected, the validation engine invokes a value extractor (see <<constraintdeclarationvalidationprocess-valueextractordefinitions>>)
```
emmanuelbernard: s/container element constraint is detected/container element constraint is constrained/

Because you don't want to described a process in a spec if you can express a behavior. For example from your sentence I could infer that I need to extract for each constraint (i.e. 2 constraints on a container element means double extraction).

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+which retrieves the value(s) from the container so they can be validated.
+This may be a single value - e.g. in the case of `Optional` which wraps exactly one value if the `Optional` is not empty -
+or multiple values in the case of collection types.
```
emmanuelbernard: Maybe s/or multiple values in the case of collection types./or multiple values in the case of `Iterable` types./ ?

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+[source, JAVA]
+----
+private Map<String, @NotEmpty List<@ValidAddress Address>> addressesByType;
+----
+====
+
+In such case multiple value extractors will be invoked.
+In the example above,
+
+* the extractor for `Map` values will be invoked to obtain all map values (lists of `Address`)
```
emmanuelbernard: I did not see the nested logic being described in the `@Valid` section.

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+* for each extracted list of addresses, the `@NotEmpty` constraint will be validated and the extractor for `List` elements will be invoked, providing the `Address` objects from each list in the map
+* the `@ValidAddress` constraint will be applied to all elements of all lists stored in the map
+
+Container element constraints must not be put to wildcard type arguments (with or without bounds), i.e. the following usage is disallowed:
```
emmanuelbernard: Should we say _disallowed_ or _not supported_?

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+@Size(max=10) private StringProperty name;
+@Positive private IntegerProperty positiveInt;
+----
+====
+
+Here the `@Size` and `@Positive` constraints don't apply to the annotated `StringProperty` and `IntegerProperty` themselves, but rather to the wrapped string and integer values, respectively.
+
+For this to work, an unambiguously resolvable value extractor must be defined which carries the `@UnwrapByDefault` annotation (see <<constraintdeclarationvalidationprocess-valueextractordefinitions-unwrapbydefault>>).
+
+If needed, the target (container or container element) of a constraint declared on a container can be explicitly specified via the `Unwrap` and `Skip` payload definitions:
```
emmanuelbernard: ah so we stayed with payload in the end :)

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+
+[source,java]
+----
+@Email(payload = Unwrapping.Unwrap.class)
+List<String> emails;
+----
+====
+
+[[constraintdeclarationvalidationprocess-valueextractordefinitions]]
+=== Value extractor definitions
```
emmanuelbernard: I think I would have put this section in Constraint Definition as it's more of a definition thingy than a declaration thingy. Have you considered it?
Maybe a top level chapter?

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+
+Validation of container element constraints as well as cascaded validation of generic container types requires access to the values stored in the container.
+The retrieval of values stored in a container is handled via implementations of the `ValueExtractor` interface:
+
+[source, JAVA, indent=0]
+----
+include::{validation-api-source-dir}javax/validation/valueextraction/ValueExtractor.java[lines=7..8;15..-1]
```
emmanuelbernard: In the JavaDoc: "Defines the logic used to extract the values to validate from an object of type {@code T}"

Should we say "to validate", after all sometimes we just cascade.

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+----
+
+The validation engine passes the container instance and a value receiver object to the `extractValues()` method.
+Value extractor implementations must invoke one of the `ValueReceiver` methods for each element contained in the container, passing the element value and, optionally, a node name.
```
emmanuelbernard: I'm a bit reluctant to the use "must" because what happens if the container is empty, must I call a method? Which one?

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+When calling
+
+* `value()`, the given value will be passed to the validation engine;
+* `iterableValue()`, the given value will be passed to the validation engine and the corresponding property path node (see <<validationapi-constraintviolation>>) will be marked as iterable, i.e. `Node#isInIterable()` returns `true`;
+* `indexedValue()`, the given value will be passed to the validation engine and the corresponding property path node will be marked as iterable and it will have set the given index, i.e. `Node#getIndex()` returns the given index value;
+* `keyedValue()`, the given value will be passed to the validation engine and the corresponding property path node will  be marked as iterable and it will have set the given key, i.e. `Node#getKey()` returns the given key value.
```
emmanuelbernard: s/will  be/will be/

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+
+When passing a non-null node name to any of the receiver methods and the extractor is called in the course of container element validation (and not during cascaded validation),
```
emmanuelbernard: s/ and the extractor/ and when the extractor/

emmanuelbernard: s/and not during cascaded validation/not during cascaded validation/

sources/constraint-declaration-validation.asciidoc
emmanuelbernard: Also we say 
"@param nodeName the name of the node to be added to the {@link Path}"

It should be more explicit
"@param nodeName the name of the node representing the container element. It will be added to the {@link Path}"

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+the map as obtained directly from the field will be passed to the map value extractor in order to obtain the map values.
+
+[[constraintdeclarationvalidationprocess-valueextractordefinitions-extractedvalue]]
+==== @ExtractedValue
+
+The `@ExtractedValue` annotation is used to denote the element extracted by a given value extractor:
+
+[source, JAVA, indent=0]
+----
+include::{validation-api-source-dir}javax/validation/valueextraction/ExtractedValue.java[lines=7..8;15..-1]
```
emmanuelbernard: In the javadoc
s/Annotation used to mark/Mark/

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+`@ExtractedValue` is to be specified on a type argument of the container type as used in the extractor definition.
+Only wildcard type arguments may be annotated with `@ExtractedValue`.
+
+[NOTE]
+====
+This implies that there may not be more than one extractor for a given generic type.
+I.e. there can be an extractor for `List<?>`, but not one for `List<String>` and one for `List<Integer>`.
+====
+
+In case an illegal value extractor definition is detected, a `ValueExtractorDefinitionException` is raised.
```
emmanuelbernard: Can we say that the exception is raised at boot time?

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+==== @UnwrapByDefault
+
+Value extractor definitions can be marked with the `@UnwrapByDefault` annotation.
+This causes constraints to be automatically applied to the wrapped value if a constraint is found for an element handled by that extractor (see <<constraintdeclarationvalidationprocess-containerelementconstraints-implicitunwrapping>>):
+
+.`@UnwrapByDefault` annotation
+====
+[source, JAVA, indent=0]
+----
+include::{validation-api-source-dir}javax/validation/valueextraction/UnwrapByDefault.java[lines=7..8;16..-1]
```
emmanuelbernard: s/Constraint/SomeConstraint/ otherwise we think it's the actual`Constraint` type

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+}
+----
+
+[[constraintdeclarationvalidationprocess-valueextractordefinitions-builtinvalueextractors]]
+==== Built-in value extractors
+
+Compatible implementations provide value extractors for the following types out of the box:
+
+* `java.util.Iterable`; `iterableValue()` must be invoked for each contained element, passing the string literal `<iterable element>` as node name
+* `java.util.List`; `indexedValue()` must be invoked for each contained element, passing the string literal `<iterable element>` as node name
```
emmanuelbernard: would it make sense to pass `<list element>` ?

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1026,6 +1100,289 @@ public class SimpleOrderService extends OrderService {
...
+* `java.util.Map`; both map keys and map values are to be supported;
+`keyedValue()` must be invoked by the map key extractor for each contained key, passing the string literal `<map key>` as node name;
+`keyedValue()` must be invoked by the map value extractor for each contained value, passing the string literal `<map value>` as node name
+* `java.util.Optional`; `value()` must be invoked if a value is present, passing `null` as node name
+* `javafx.beans.observable.ObservableValue`; `value()` must be invoked with the observable value, passing `null` as node name; the extractor must be marked with `@UnwrapByDefault`
+* `javafx.beans.property.ReadOnlyListProperty` and `javafx.beans.property.ListProperty`; `indexedValue()` must be invoked for each contained element, passing the string literal `<iterable element>` as node name
```
emmanuelbernard: same here

emmanuelbernard:
@gunnarmorling first pass

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1573,7 +1930,184 @@ The resolutions shown in <<table-constraintvalidator-resolution>> occur.
...
+Value extractors can be registered with the validation engine in the following ways (in increasing order of priority):
+
+* Provided by the validation engine itself (see <<constraintdeclarationvalidationprocess-valueextractordefinitions-builtinvalueextractors>>)
+* Via the Java https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html[service loader] mechanism; for this the file _META-INF/services/javax.validation.valueextraction.ValueExtractor_ must be provided,
+with the fully-qualified name(s) of one or more extractor implementations as its contents
+* By specifying the fully-qualified class name of one or several extractors in `META-INF/validation.xml` (see <<validationapi-bootstrapping-xmlconfiguration>>)
+* By invoking the method `Configuration#addValueExtractor(ValueExtractor<?>)` (to apply it at the validation factory level)
+* By invoking the method `ValidatorContext#addValueExtractor(ValueExtractor<?>)` (to apply it for a single `Validator`)
+
+A value extractor for a given type and type parameter specified at a higher level overrides any other extractors for the same type and type parameter given at lower levels.
```
emmanuelbernard: s/levels/priorities/g

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1573,7 +1930,184 @@ The resolutions shown in <<table-constraintvalidator-resolution>> occur.
...
+[[constraintdeclarationvalidationprocess-validationroutine-valueextractorresolution-algorithm-constraints]]
+===== ValueExtractor resolution algorithm for container element constraints
+
+For a container with the declared type `C` whose element type is hosting a constraint, the following resolution rules apply for identifying the value extractor:
+
+* A `ValueExtractor<T>` is said to be _type-compliant_ with `C`, if `C` is a subtype of `T`
+(according to the http://java.sun.com/docs/books/jls/third_edition/html/typesValues.html#4.10[Java Language Specification 3rd edition chapter 4.10 Subtyping]).
+Note that `C` is a subtype of `T` if `C` = `T`.
+* A `ValueExtractor` implementation is said to be _container-element-compliant_ with `C`, if `C` is a generic container type and the value extractor implementation handles a type parameter that maps to the constrained type argument.
+* If no `ValueExtractor` type-compliant and container-element-compliant with `C` is found among the available value extractors, a `ConstraintDeclarationException` is raised.
```
emmanuelbernard: Interesting, you've thought this through and an exception is better than ignoring? Just asking here to make sure we are not missing anything.

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1573,7 +1930,184 @@ The resolutions shown in <<table-constraintvalidator-resolution>> occur.
...
+* A `ValueExtractor<U>` type-compliant with `C'` is considered _strictly more specific_ than a `ValueExtractor<V>` compliant with `C'` if `U` is a strict subtype of `V`.
+`U` is a strict subtype of `V` if `U` is a subtype of `V` and `U` != `V`.
+* A `ValueExtractor<U>` type-compliant with `C'` is considered maximally specific if no other `ValueExtractor<V>` type-compliant with `C'` is strictly more specific than `ValueExtractor<U>`.
+* If more than one maximally specific and container-element-compliant `ValueExtractor` is found, a `ConstraintDeclarationException` is raised.
+
+[NOTE]
+.Implementation note
+====
+Extractor retrieval for container element constraints is based on the runtime type of the elements marked with `@Valid`
+(this is to be consistent with the semantics of property paths for cascaded validation defined in Bean Validation 1.x),
```
emmanuelbernard: People reading the spec don't care that things are from 1.x. so s/ defined in Bean Validation 1.x//

sources/constraint-declaration-validation.asciidoc
```diff
@@ -1573,7 +1930,184 @@ The resolutions shown in <<table-constraintvalidator-resolution>> occur.
...
+hence it is not possible to perform the extractor resolution only once and cache the result.
```
emmanuelbernard: couldn't you keep a hot cache per type and look it up first? It would not be caching the resolution per site but it would across a `ValidatorFactory`

emmanuelbernard: If I'm correct, I think you should soften the "recommended" tone

sources/constraint-metadata.asciidoc
```diff
@@ -382,8 +382,33 @@ This interface is returned by [methodname]`MethodDescriptor.getReturnValueDescri
...
+[[constraintmetadata-containerdescriptor]]
+=== ContainerDescriptor and ContainerElementTypeDescriptor
+
+The `ContainerDescriptor` interface describes those elements that can host container element constraints (see <<constraintdeclarationvalidationprocess-containerelementconstraints>>).
+
+`ContainerDescriptor` is extended by `PropertyDescriptor`, `ParameterDescriptor`, `ReturnValueDescriptor` and `ContainerElementTypeDescriptor`.
+
+[source, JAVA, indent=0]
+----
+include::{validation-api-source-dir}javax/validation/metadata/ContainerDescriptor.java[lines=7..8;13..-1]
```
emmanuelbernard: I my rendering of the spec I have

```
Unresolved directive in sources/constraint-metadata.asciidoc - include::../target/validation-api/javax/validation/metadata/ContainerDescriptor.java[lines=7..8;13..-1]
```

sources/constraint-metadata.asciidoc
```diff
@@ -382,8 +382,33 @@ This interface is returned by [methodname]`MethodDescriptor.getReturnValueDescri
...
+----
+
+`getConstraintsForContainerElementType()` returns a descriptor representing the constraints of the specified type argument.
+`getConstrainedContainerElementTypes()` returns descriptors for all those type arguments that are hosting at least one constraint or are marked for cascaded validation.
```
emmanuelbernard: Interesting, do we also do that for properties, we don't return them if they are not constrained?

sources/exception.asciidoc
```diff
@@ -141,6 +141,33 @@ These exception cases can be determined at compile time by a tool such as an ann
 These exception cases can be determined at compile time by a tool such as an annotation processor.
 ====
 
+[[exception-valueextractordefinition]]
```
emmanuelbernard: I think the exception chapter should say that when no VE is found on a constrained container element, exception X is raised.
Same for a container type that leads to two maximizing stuff VE

sources/validation-api.asciidoc
```diff
@@ -1823,6 +1840,7 @@ The responsibility of the [classname]`Configuration` is to collect configuration
 * the constraint validator factory instance
 * the parameter name provider instance
 * the clock provider instance
+* one or more value extractor instance(s)
```
emmanuelbernard: you can define zero of them as well.

sources/validation-api.asciidoc
```diff
@@ -2280,12 +2302,11 @@ include::{validation-api-source-dir}javax/validation/bootstrap/ProviderSpecificB
 * [methodname]`public static <T extends Configuration<T>, U extends ValidationProvider<T>> ProviderSpecificBootstrap<T> byProvider(Class<U> providerType)`
 
 
-The bootstrap API is designed to allow complete portability amongst Bean Validation provider implementations. [tck-testable]#The bootstrap implementation must ensure it can bootstrap third party providers.#
+The bootstrap API is designed to allow complete portability among Bean Validation provider implementations. [tck-testable]#The bootstrap implementation must ensure it can bootstrap third party providers.#
```
emmanuelbernard: Who's the `among` nazi? My money is on @gsmet

sources/constraint-declaration-validation.asciidoc
gunnarmorling: Good idea.

sources/constraint-declaration-validation.asciidoc
gunnarmorling: Changed to

> Container element constraints can be used within the following declarations:

sources/constraint-declaration-validation.asciidoc
gunnarmorling: Changed it to 

> When a field, property, executable parameter or method return value which is of a container type gets validated, then all values contained in the container will be validated provided that their container element is constrained.

sources/constraint-declaration-validation.asciidoc
gunnarmorling: It could be maps, too. Hence I went for the generic term "collection" (not `Collection`).

sources/constraint-declaration-validation.asciidoc
gunnarmorling: Which section do you mean specifically?

sources/constraint-declaration-validation.asciidoc
gunnarmorling: I changed it to "unsupported", although the only thing theoretically making sense is an extends bound (`? extends String`), as otherwise no constraint validator can be resolved.

sources/constraint-declaration-validation.asciidoc
gunnarmorling: Yeah, seems like the best we can do.

sources/constraint-declaration-validation.asciidoc
gunnarmorling: I considered it, but in "Constraint Definition" it seemed wrong. We can do a separate chapter, although it seems a bit heavy.

sources/constraint-declaration-validation.asciidoc
gunnarmorling: +1

sources/constraint-declaration-validation.asciidoc
gunnarmorling: It says "must invoke ... for each element contained". I.e. if there is no element, no method is to be invoked.

sources/constraint-declaration-validation.asciidoc
gunnarmorling: I'd rather leave it open. Perhaps a provider processes service loader discovered extractors lazily?

sources/constraint-declaration-validation.asciidoc
gunnarmorling: Changed; worse, this didn't use the payload() style. Changed it.

sources/constraint-declaration-validation.asciidoc
gunnarmorling: Agh, `<iterable element>` is a glitch indeed. Could swear it was `<list element>` at some point.

sources/constraint-declaration-validation.asciidoc
gunnarmorling: Yes, I think an exception is better. Otherwise it's like ignoring a constraint if no validator is found. This shouldn't go unnoticed.

sources/constraint-declaration-validation.asciidoc
gunnarmorling: Good thinking; I've removed that last sentence.

sources/constraint-metadata.asciidoc
gunnarmorling: I think you saw that because you didn't have the latest API sources.

sources/constraint-metadata.asciidoc
gunnarmorling: Yes, see `BeanDescriptor#getConstrainedProperties()`. Although I'm wondering whether it's better to do it as for executable parameters, which we all return in the order of their declaration.

sources/validation-api.asciidoc
gunnarmorling: I'll only say that it wasn't me ;)

sources/exception.asciidoc
gunnarmorling: Added.

sources/constraint-declaration-validation.asciidoc
emmanuelbernard: The question is whether that's desirable to have such behavior for a user.

emmanuelbernard: but we can remain vague. Better be cautious.

sources/constraint-metadata.asciidoc
emmanuelbernard: We had that discussion back in the 1.0 days, but I'm not sure whether we had a strong reason.

gunnarmorling:
Thinking about it (see bv-dev mail), this seems more in the category of
method parameters where we also return descriptors for all of them. Going
to change it in that spirit.

PR #156, opened 2017-04-21
BVAL-596 Move ValidateUnwrappedValue to javax.validation.metadata

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-596

### DIFF
--- a/sources/constraint-metadata.asciidoc
+++ b/sources/constraint-metadata.asciidoc
@@ -436,6 +436,13 @@ include::{validation-api-source-dir}javax/validation/metadata/ConstraintDescript
 
 [tck-testable]#[methodname]`getConstraintValidatorClasses()` returns the [classname]`ConstraintValidator` classes associated with the constraint.#
 
+[tck-testable]#[methodname]`validateUnwrappedValue()` returns a [classname]`ValidateUnwrappedValue` instance describing the unwrapping behavior.#
+
+[source, JAVA, indent=0]
+----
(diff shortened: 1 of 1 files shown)


gunnarmorling:
Isn't there any include of that file that needs adapting?

gsmet:
@gunnarmorling nope, it's not described in the spec AFAICS. Maybe it should be?

gunnarmorling:
Yes, absolutely. Each API type should be contained in the spec.

gsmet:
@gunnarmorling I added one more commit to mention `validateUnwrappedValue()`.

sources/constraint-metadata.asciidoc
```diff
@@ -436,6 +436,13 @@ include::{validation-api-source-dir}javax/validation/metadata/ConstraintDescript
 
 [tck-testable]#[methodname]`getConstraintValidatorClasses()` returns the [classname]`ConstraintValidator` classes associated with the constraint.#
 
+[tck-testable]#[methodname]`validateUnwrappedValue()` returns a [classname]`ValidateUnwrappedValue` value describing the unwrapping behavior.#
```
gunnarmorling: perhaps s/value describing/instance describing/?

gsmet: Ah, yes, it's better. Force pushed.

gunnarmorling:
Thanks, @gsmet!

PR #159, opened 2017-04-23
BVAL-591 follow-up

gunnarmorling:


sources/constraint-declaration-validation.asciidoc
```diff
@@ -166,6 +165,22 @@ public class User {
...
+[source, JAVA]
+----
+public class User {
+    private Map<String, @Valid List<@Valid Address>> addressesByType;
+}
+----
+====
+
+In the last example, all `Address` objects contained in the lists of the `addressesByType` map will be validated.
+Two value extractor are invoked for this:
```
gsmet: extractors

sources/constraint-declaration-validation.asciidoc
```diff
@@ -166,6 +165,22 @@ public class User {
 ----
 ====
 
+.Making a nested collection subject to graph validation
+====
+[source, JAVA]
+----
+public class User {
+    private Map<String, @Valid List<@Valid Address>> addressesByType;
```
gsmet: I don't think we need the `@Valid` on the List or do we?

gunnarmorling: That's the question essentially; I've filed [BVAL-601](https://hibernate.atlassian.net/browse/BVAL-601) to clarify it (last question in there). For now I'd prefer to stay safe and require them all along the path.

sources/constraint-declaration-validation.asciidoc
gunnarmorling: Fixed and force-pushed.

gsmet:
Merged, thanks!

PR #166, opened 2017-04-26
BVAL-610 Inject the section id constants in tck-audit.xml

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-610


gunnarmorling:
The old XSL guru fighting back ;)

PR #168, opened 2017-05-06
BVAL-614 - BVAL-615 -  BVAL-526 Asciidoctor work

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-614
* https://hibernate.atlassian.net/browse/BVAL-615
* https://hibernate.atlassian.net/browse/BVAL-526

Based on:
 * https://github.com/asciidoctor/asciidoctor-ant/pull/60
 * https://github.com/hibernate/hibernate-asciidoctor-extensions/pull/2

@marko-bekhta thought you might be interested in this one. It should solve the issue we have with the extensions and make it possible to upgrade more easily to new versions of Asciidoctorj.

It also removes quite a lot of boilerplate code by using the new if/unless feature introduced in Ant 1.9.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -14,9 +14,13 @@
 	<property name="bv.version.qualifier" value="" /><!-- e.g. ' (Early Draft 1)' - be careful about the space before the qualifier -->
 	<property name="bv.revdate" value="${current.date}" /> <!-- Replace with fixed date when releasing -->
 	<property name="license" value="license-evaluation" />
-	<property name="asciidoctor-ant.version" value="1.5.3" />
+	<property name="asciidoctor-ant.version" value="1.6.0-alpha.3" />
+	<property name="asciidoctorj.version" value="1.6.0-alpha.3" />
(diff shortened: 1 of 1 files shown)


marko-bekhta:
@gsmet thanks for the notification! I actually saw that issue about "uberjar" in asciidoctor-ant and waited for the release to see how it'll work. But you've already done it :)

As for the numbering - if that's not a desired behavior, you've seen the extension - it can easily be changed to use same sequence for tables and examples. Just let me know I'll make a change. Or if you'd like to do it on your own - that's fine as well :)

gsmet:
@gunnarmorling this one can be merged now. Note that it will introduce a lot of changes in the HTML diff but it's now or never I think.

@marko-bekhta thanks for your hard work on the Asciidoctor front for this one. It wasn't an easy one.

BTW, one thing I noticed is that we are not very consistent when including BV APIs in the spec. Sometimes they are wrapped in an example and sometimes not.

gunnarmorling:
Rebased and applied. Thanks a lot @marko-bekhta and @gsmet, that was a tough one :)

PR #169, opened 2017-05-09
BVAL-603 Clarify that value extractors are subject to CDI, too

gunnarmorling:
https://hibernate.atlassian.net/browse/BVAL-603


gsmet:
Rebased and merged, thanks!

PR #173, opened 2017-05-16
BVAL-620 Discouraging usage of old and new container validation style at the same time

gunnarmorling:
https://hibernate.atlassian.net/browse/BVAL-620


gsmet:
Merged, thanks!

PR #174, opened 2017-05-16
BVAL-517 Adding appendix with recommended module name

gunnarmorling:
https://hibernate.atlassian.net/browse/BVAL-517


gsmet:
Looks good. Should we wait for Emmanuel?

gunnarmorling:
No, I think we can merge it. We agreed on it when discussing it last week. Thanks!

PR #175, opened 2017-05-16
BVAL-592 Adding inContainer(Class, int) to node builder API

gunnarmorling:
https://hibernate.atlassian.net/browse/BVAL-592


gsmet:
Merged, thanks!

PR #176, opened 2017-05-16
BVAL-579 Update source include of ExtractedValue

gsmet:
This is just to update the source include of ExtractedValue so that we don't forget about it.

We need some additional wording for this feature but I thought you might prefer doing it?


gunnarmorling:
Merged; the corresponding spec change is done here: https://github.com/beanvalidation/beanvalidation-spec/pull/177

PR #182, opened 2017-06-06
BVAL-635 Adding current git revision to generated audit file

gunnarmorling:
https://hibernate.atlassian.net/browse/BVAL-635


gsmet:
I end up having the following output in the tck-audit.xml file:

```
<!-- Generated by tck-audit.xsl at 2017-06-06 17:32  (revision )2.0.0.Beta1-33-g497a5df -->
```

Do you also see this issue?

It looks more like a bug in the XSLT processor but we should try to have it right.

BTW, is there a specific reason while you concat with a space?

gunnarmorling:
@gsmet Re-worked that commit. I've worked around that issue by adding a separate xsl:text element now. On the concatenation, I'm not sure, it was C&P from the date line, there I reckon we did it to have an extra-space before the closing "-->", probably to work around that same issue. Looks good to me now.

gunnarmorling:
Btw. you posted as an output "2.0.0.Beta1-33-g497a5df" - could it be you haven't fetched upstream since the last release? Because it should show the latest tag there, so that's what I'm getting: "2.0.0.Beta2-5-g8d348a2".

The format is "<latest-tag>-<#no.-of-commits-since-then>-g<commit-sha>". And as I learned, the "g" stands for "git" which supposedly should help when dealing with revision markers from multiple SCMs. Go figure.

gsmet:
Merged, thanks!

PR #184, opened 2017-06-07
BVAL-636 Mandating JavaFX extractors only in environments where JavaF…

gunnarmorling:
…X is present

https://hibernate.atlassian.net/browse/BVAL-636


gsmet:
So, this makes the Oracle JDK a requirement to validate the TCK as we will need an environment where JavaFX is available to test this.

Agreed?

gunnarmorling:
It makes JavaFX a requirement for running and passing the TCK. I think one
can install OpenJFX next to OpenJDK, so technically it doesn't require the
Oracle JDK. I've filed a BVTCK for keeping track of it.

gsmet:
Merged, thanks!

PR #188, opened 2017-06-08
BVAL-624 Value extractors discovered via the service loader are CDI-e…

gunnarmorling:
…nabled

https://hibernate.atlassian.net/browse/BVAL-624

@gsmet I think it's good, there's one difference to the XML case, though: the SL returns ready-made instances of the extractors, i.e. we're not in charge of instantiation. I don't think that's a problem, though. We should know once implementing and TCK-ing it.


gsmet:
Fixed a typo (`applies for` -> `applies to`) and merged. Thanks!

PR #190, opened 2017-06-13
BVAL-646 Clarifications around configuration of container element typ…

gunnarmorling:
…es in XML

https://hibernate.atlassian.net/browse/BVAL-646


gsmet:
Perfect, thanks!

PR #191, opened 2017-06-13
BVAL-637 Clarifying extracted value type for optional unwrappers

gunnarmorling:
https://hibernate.atlassian.net/browse/BVAL-637


gsmet:
Rebased and merged, thanks!

PR #193, opened 2017-06-15
BVAL-601 Clarifying usage of @Valid for nested generic containers

gunnarmorling:
https://hibernate.atlassian.net/browse/BVAL-601


gsmet:
Merged, thanks!

PR #195, opened 2017-06-15
BVAL-648 Adding OptionalInt example to value extractor resolution cha…

gunnarmorling:
…pter

https://hibernate.atlassian.net/browse/BVAL-648


gsmet:
Merged, thanks!

PR #196, opened 2017-06-15
BVAL-649 Updating examples showing parameter names to use the actual …

gunnarmorling:
…parameter name assuming it's available

https://hibernate.atlassian.net/browse/BVAL-649


gsmet:
Fixed a few typos (`can` was repeated several time in some comments) and merged.

PR #198, opened 2017-06-15
BVAL-652 Clarifying behavior in case several value extractors for the…

gunnarmorling:
… same type and type parameter are given

https://hibernate.atlassian.net/browse/BVAL-652


gsmet:
Merged, thanks!

PR #201, opened 2017-06-16
BVAL-657 Fixing wrong map key in constraint validator context example

gunnarmorling:
https://hibernate.atlassian.net/projects/BVAL/issues/BVAL-657


gsmet:
Merged!

PR #202, opened 2017-06-16
BVAL-552 inIterable() should return true for arrays, too

gunnarmorling:
https://hibernate.atlassian.net/browse/BVAL-552


gsmet:
Merged, thanks!

PR #208, opened 2017-06-20
BVAL-552 Describing behavior of node methods more generically, taking…

gunnarmorling:
… custom container types into account


gsmet:
Merged, thanks!

PR #210, opened 2017-06-21
BVAL-666 Making explicit that container element constraints and @Vali…

gunnarmorling:
…d are not supported for type parameters and within type definitions

https://hibernate.atlassian.net/projects/BVAL/issues/BVAL-666

### DIFF
--- a/sources/constraint-declaration-validation.asciidoc
+++ b/sources/constraint-declaration-validation.asciidoc
@@ -125,6 +125,10 @@ but not both (in order to prevent the container elements from being validated tw
 
 [tck-testable]#The [classname]`@Valid` annotation is applied recursively.# A conforming implementation avoids infinite loops according to the rules described in <<constraintdeclarationvalidationprocess-validationroutine-graphvalidation>>.
 
+[tck-not-testable]#It is not supported to put `@Valid` to the type parameters of generic types or methods.#
++[tck-not-testable]#It is also not supported to put `@Valid` to type arguments within the `extends` or `implements` clauses of type definitions.#
+A future revision of this specification might define support for such usages of `@Valid`.
+
 [[constraintdeclarationvalidationprocess-requirements-graphvalidation-examples]]
 ===== Examples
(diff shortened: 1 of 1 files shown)


sources/constraint-declaration-validation.asciidoc
```diff
@@ -125,6 +125,10 @@ but not both (in order to prevent the container elements from being validated tw
 
 [tck-testable]#The [classname]`@Valid` annotation is applied recursively.# A conforming implementation avoids infinite loops according to the rules described in <<constraintdeclarationvalidationprocess-validationroutine-graphvalidation>>.
 
+[tck-not-testable]#It is not supported to put `@Valid` to the type parameters of generic types or methods.#
++[tck-not-testable]#It is also not supported to put `@Valid` to type arguments within the `extends` or `implements` clauses of type definitions.#
```
gsmet: There's a + here at the beginning of the line. Will fix it and merge.

gsmet:
Merged, thanks!

sources/constraint-declaration-validation.asciidoc
```diff
@@ -841,16 +845,30 @@ In the example above,
...
 ====
 [source, JAVA]
 ----
-private List<@NotEmpty ?> list;
+public class NonNullList<@NotNull T> {
+    [...]
+}
+
+public class ContainerFactory {
+    <@NotNull T> Container<T> instantiateContainer(Class<T> clazz) { [...] }
```
mbenson: I would point out here that Class is perhaps a bad example as there isn't any meaningful way to extract instances of its type parameter.

gunnarmorling:
How should it be changed, just passing an instance of T for the sake of the
example?

mbenson:
That sounds better than Class, and is easy too.

Matt

gunnarmorling:
Ok, cool. Sent https://github.com/beanvalidation/beanvalidation-spec/pull/214.

PR #216, opened 2017-06-21
BVAL-661 Add type parameter and type argument definitions from our glossary

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-661

### DIFF
--- a/sources/terminology.asciidoc
+++ b/sources/terminology.asciidoc
@@ -32,5 +32,16 @@ Getter:: Method whose:
 * name starts with `is`, has no parameter and is returning [classname]`boolean`
 Generic type:: A class or interface with one or more type parameters, e.g. `class List<E> { ... }`
 Parameterized type (instantiated type):: A type created from a generic type by providing an actual type argument per formal type parameter, e.g. `List<String>`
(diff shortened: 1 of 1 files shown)


sources/terminology.asciidoc
```diff
@@ -32,5 +32,17 @@ Getter:: Method whose:
 * name starts with `is`, has no parameter and is returning [classname]`boolean`
 Generic type:: A class or interface with one or more type parameters, e.g. `class List<E> { ... }`
 Parameterized type (instantiated type):: A type created from a generic type by providing an actual type argument per formal type parameter, e.g. `List<String>`
+Type argument:: A reference type or a wildcard that is used for instantiation / invocation of a generic type or a reference type used for instantiation / invocation of a generic method
```
gunnarmorling: I didn't add it because I found the examples not very helpful: what is the type argument here exactly. We should at least say that it is the `?` and `String` here. Same for the othe one.

gsmet: Force pushed some additional info that should make it clear.

gunnarmorling:
Merged; thanks, @gsmet!

PR #221, opened 2017-06-23
BVAL-676 Differentiate the API listings from the examples

gsmet:
* https://hibernate.atlassian.net/browse/BVAL-676

@gunnarmorling I added backticks in the title example. Not sure if I like the result. Opinion welcome, I'll get rid of them if you don't like them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,8 @@ bin/
 
 #Edit
 *.*~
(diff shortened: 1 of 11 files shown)


gunnarmorling:
Awesome! Thanks a lot, @gsmet.

PR #225, opened 2017-06-29
BVAL-682 Clarifying semantics of getTypeArgumentIndex() in edge case …

gunnarmorling:
…of extractor for non-generic container appending a node


gsmet:
Merged, thanks!

PR #227, opened 2017-07-03
BVAL-674 Adapting to changed method name; adding unwrap() method to spec

gunnarmorling:
https://hibernate.atlassian.net/projects/BVAL/issues/BVAL-674


gsmet:
Merged, thanks!

PR #229, opened 2017-07-04
BVAL-687 Adding final license

gunnarmorling:
@gsmet, let's merge it once the CR 2 is out.

This won't be added to the rendered spec document itself (see the description of [https://hibernate.atlassian.net/browse/BVAL-687](BVAL-687)), but I'm adding it here for the sake of completeness and also so to have it handy when doing the final submission.


gsmet:
Merged!

PR #230, opened 2017-07-04
BVAL-667 Replacing @NotEmpty example constraint with @NonEmpty to avo…

gunnarmorling:
…id confusion with the built-in constraint of the same name;

Specifically, some of the example assertions rely on the fact that the constraint is a composed constraint, which isn't the case for the built-in @NotEmpty constraint

https://hibernate.atlassian.net/browse/BVAL-667


gunnarmorling:
We have two definitions of the example `@NonEmpty` constraint, which are the same essentially. So we could share the same definition, but then it seems good to show the definition again in the context of the second chapter using it, so to avoid a back reference to the original definition given very far away. Thoughts?

gsmet:
Agreed and merged. Thanks!

PR #232, opened 2017-07-04
BVAL-604 Avoiding forced line wraps

gunnarmorling:
https://hibernate.atlassian.net/browse/BVAL-604


gunnarmorling:
@gsmet Pushed one more commit for the change log. Not quite sure how to handle it for future releases. Seems doable but not too easy to adapt the release script accordlingy? WDYT?

gsmet:
Merged.

PR #235, opened 2017-07-05
BVAL-682 Misc. clarifications around container element validation

gunnarmorling:
https://hibernate.atlassian.net/projects/BVAL/issues/BVAL-682

### DIFF
--- a/sources/exception.asciidoc
+++ b/sources/exception.asciidoc
@@ -157,14 +157,7 @@ Reasons for raising this exception include:
 * The extracted type is not marked with `@ExtractedValue`
 * The `@ExtractedValue` annotation is given more than once for one value extractor type
 
-[role="api"]
-.`ValueExtractorDefinitionException` exception
-====
-[source,java]
-----
-include::{validation-api-source-dir}javax/validation/valueextraction/ValueExtractorDefinitionException.java[lines=7..8;11..-1]
-----
-====
+`ValueExtractorDefinitionException` is a subclass of `ValidationException`.
 
 [[exception-valueextractordeclaration]]
 === Value extractor declaration: [classname]`ValueExtractorDeclarationException`
@@ -172,14 +165,7 @@ include::{validation-api-source-dir}javax/validation/valueextraction/ValueExtrac
 [tck-testable]#When detecting an illegal configuration of value extractors, a `ValueExtractorDeclarationException` will be raised.#
 One example is the configuration of multiple extractors for the same container element type in `META-INF/validation.xml`.
 
-[role="api"]
-.`ValueExtractorDeclarationException` exception
-====
-[source,java]
-----
-include::{validation-api-source-dir}javax/validation/valueextraction/ValueExtractorDeclarationException.java[lines=7..8;11..-1]
-----
-====
+`ValueExtractorDeclarationException` is a subclass of `ValidationException`.
 
 [[exception-noprovider]]
 === No Bean Validation Provider detected: `NoProviderFoundException`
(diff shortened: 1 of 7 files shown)


sources/exception.asciidoc
```diff
@@ -157,29 +157,15 @@ Reasons for raising this exception include:
 * The extracted type is not marked with `@ExtractedValue`
 * The `@ExtractedValue` annotation is given more than once for one value extractor type
 
-[role="api"]
```
gunnarmorling: I've removed the includes to be consistent with what we're doing for the other exceptions.

@gsmet Should we sprinkle some tck-testable markers over the examples given for exception cases?

gsmet: I don't think it's a good idea. They are only examples. The tck-testable assertions should be present in the according section.

sources/validation-api.asciidoc
```diff
@@ -529,7 +529,7 @@ It is possible to narrow a node instance to its precise type and extract node sp
 [tck-testable]
 --
 `BeanNode`, `PropertyNode` and `ContainerElementNode` host `getContainerClass()` and `getTypeArgumentIndex()`.
-If the node represents an element that is contained in a generic container such as `Optional`, `List` or `Map`,
+If the node represents an element that is contained in a container such as `Optional`, `List` or `Map`,
 the former returns the declared type of the container and the latter returns the index of the affected type argument.
```
gsmet: Why did you remove the `generic` word? We are talking about the type argument so it only makes sense for generic containers.

sources/value-extractor-definition.asciidoc
```diff
@@ -106,8 +110,8 @@ Compatible implementations provide value extractors for the following types out
...
-* [tck-testable]#`java.util.Optional`; `value()` must be invoked (passing the contained value or `null` if none is present), passing `null` as node name#
-* [tck-testable]#`java.util.OptionalInt`, `java.util.OptionalLong` and `java.util.OptionalDouble`; the extracted value types must be `java.lang.Integer`, `java.lang.Long` and `java.lang.Double`, respectively. `value()` must be invoked (passing the contained value or `null` if none is present), passing `null` as node name. The extractors must be marked with `@UnwrapByDefault`.#
+* [tck-testable]#`java.util.Optional`; `value()` must be invoked, passing `null` as node name and passing the contained object as value or `null` if none is present)#
+* [tck-testable]#`java.util.OptionalInt`, `java.util.OptionalLong` and `java.util.OptionalDouble`; the extracted value types must be `java.lang.Integer`, `java.lang.Long` and `java.lang.Double`, respectively. `value()` must be invoked, passing `null` as node name and passing the contained number as value or `null` if none is present). The extractors must be marked with `@UnwrapByDefault`.#
```
gsmet: `present)` I can't find the starting parenthesis so I suppose we should remove it?

sources/value-extractor-definition.asciidoc
```diff
@@ -106,8 +110,8 @@ Compatible implementations provide value extractors for the following types out
 * [tck-testable]#`java.util.Map`; both map keys and map values are to be supported;
 `keyedValue()` must be invoked by the map key extractor for each contained key, passing the string literal `<map key>` as node name;
 `keyedValue()` must be invoked by the map value extractor for each contained value, passing the string literal `<map value>` as node name#
-* [tck-testable]#`java.util.Optional`; `value()` must be invoked (passing the contained value or `null` if none is present), passing `null` as node name#
-* [tck-testable]#`java.util.OptionalInt`, `java.util.OptionalLong` and `java.util.OptionalDouble`; the extracted value types must be `java.lang.Integer`, `java.lang.Long` and `java.lang.Double`, respectively. `value()` must be invoked (passing the contained value or `null` if none is present), passing `null` as node name. The extractors must be marked with `@UnwrapByDefault`.#
+* [tck-testable]#`java.util.Optional`; `value()` must be invoked, passing `null` as node name and passing the contained object as value or `null` if none is present)#
```
gsmet: present) I can't find the starting parenthesis so I suppose we should remove it?

sources/exception.asciidoc
gunnarmorling: Ok. Note these are not "examples" in the usual sense, but some of the specific cases where this exception is to be thrown. But I think it's covered in the specific sections, too.

sources/validation-api.asciidoc
gunnarmorling: I removed it to cover the case where `getContainerClass()` returns the container class for a non-generic container. `getTypeArgumentIndex()` returns null in that case as per its documentation. I'll add "the index of the affected type argument if the container is of a generic type".

sources/value-extractor-definition.asciidoc
gunnarmorling: Good spot.

gunnarmorling:
@gsmet thanks a lot for your remarks - reworked accordingly and force-pushed the commit.

PR #240, opened 2017-07-11
BVAL-695 Addressing misc. findings from review

gunnarmorling:
https://hibernate.atlassian.net/browse/BVAL-695

### DIFF
--- a/sources/value-extractor-definition.asciidoc
+++ b/sources/value-extractor-definition.asciidoc
@@ -160,8 +160,8 @@ public class OptionalValueExtractor implements ValueExtractor<Optional<@Extracte
 
 This extractor passes the element wrapped by the given `Optional` to the receiver object, if present.
 `null` is passed as a node name, causing no node to be appended to the resulting property path.
-I.e. when the `@Size` constraint in `private Optional<@Size(min=1) String name;` is violated,
-the resulting property path will be the same as if a constraint hosted on the `name` field itself was violated.
+I.e. when the `@Size` constraint in `Optional<@Size(min=1) String getName() { ... }` is violated,
+the resulting property path will be the same as if a constraint hosted on the `getName` property itself was violated.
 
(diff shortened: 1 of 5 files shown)


spec-examples/src/test/java/org/beanvalidation/specexamples/constraintmetadata/EmployeeImpl.java
```diff
@@ -32,7 +29,7 @@ public Roles getRoles() {
 	}
 
 	@Override
-	public Map<@NotNull String, @Valid Account> getAccounts() {
```
gunnarmorling: Would be nice to double-check on this - they seemed redundant to me as they are already given on the overridden super-type method.

sources/value-extractor-definition.asciidoc
```diff
@@ -160,8 +160,8 @@ public class OptionalValueExtractor implements ValueExtractor<Optional<@Extracte
 
 This extractor passes the element wrapped by the given `Optional` to the receiver object, if present.
 `null` is passed as a node name, causing no node to be appended to the resulting property path.
-I.e. when the `@Size` constraint in `private Optional<@Size(min=1) String name;` is violated,
-the resulting property path will be the same as if a constraint hosted on the `name` field itself was violated.
+I.e. when the `@Size` constraint in `Optional<@Size(min=1) String getName() { ... }` is violated,
```
gsmet: Mmmh `getName` property feels weird. Wouldn't it be the `name` property anyway?

spec-examples/src/test/java/org/beanvalidation/specexamples/constraintmetadata/EmployeeImpl.java
gsmet: Looks good.

gsmet:
Amended and merged, thanks!

PR #243, opened 2019-07-11
BVAL-730 Change header to mention Eclipse Foundation License Specifation

emmanuelbernard:
Remove ASL license


spec-examples/src/test/java/org/beanvalidation/specexamples/constraintdeclarationvalidation/resolution/NonEmpty.java
```diff
@@ -1,8 +1,8 @@
 /*
- * Bean Validation: constrain once, validate everywhere.
+ * Jakarta Bean Validation: constrain once, validate everywhere.
  *
- * License: Apache License, Version 2.0
```
gunnarmorling: @emmanuelbernard, so far I understood the spec and API sources would remain under Apache v2 license, see also the latest mail from @starksm64. But it still doesn't quite compute to me, should the sources be Apache v2 but the published JavaDoc EFSL, Scott?

gsmet: Yeah, that's what I asked on Zulip and apparently, Emmanuel agrees.

gsmet: Isn't it the exact same situation as before i.e. the final versions of spec and Javadoc has to be under the Eclipse license (they were using an Oracle one before)?

And the API is under ASL 2 and stays ASL 2, even in Final.

emmanuelbernard: I got confirmation from Scott that the previous situation is similar. Just put the EFSL in the final build of a spec.

sources/xml-descriptor.asciidoc
```diff
@@ -757,9 +757,9 @@ From Bean Validation revision 1.1 onwards, mapping authors must specify the used
 
     <xs:annotation>
         <xs:documentation><![CDATA[
-            This is the XML Schema for Bean Validation constraint mapping files.
```
gunnarmorling: So this would need similar doing in the actual XSD files themselves.

gsmet:
FWIW I plan to review it today.

sources/xml-descriptor.asciidoc
gunnarmorling: Ok, that's handled already in https://github.com/beanvalidation/beanvalidation-api/pull/148/.

emmanuelbernard:
Good for review @gsmet

sources/validation-api.asciidoc
```diff
@@ -1321,7 +1321,7 @@ The locale to be used for message interpolation is defined as following:
 
 ===== Message expressions using {spec-name-expressionlanguage}
 
-[tck-testable]#The default message interpolation allows the use of {spec-name-expressionlanguage}.# [tck-testable]#Expressions to be evaluated by {spec-name-expressionlanguage}need to be enclosed in `${}` within the message descriptor# . The following properties and beans have to be made available in the {spec-name-expressionlanguage}context:
+[tck-testable]#The default message interpolation allows the use of {spec-name-expressionlanguage}.# [tck-testable]#Expressions to be evaluated by {spec-name-expressionlanguage} need to be enclosed in `${}` within the message descriptor# . The following properties and beans have to be made available in the {spec-name-expressionlanguage} context:
```
gsmet: This one fixes some spaces issue. It should be squashed as a fixup commit.

sources/introduction.asciidoc
```diff
@@ -7,7 +7,7 @@
 
 == Introduction
 
-This document is the specification of the Java API for Java{spec-name-bv} in {spec-name-eeplatform} and Java SE. The technical objective of this work is to provide an object level constraint declaration and validation facility for the Java application developer, as well as a constraint metadata repository and query API.
+This document is the specification of the Java API for JavaBean validation in {spec-name-eeplatform} and Java SE. The technical objective of this work is to provide an object level constraint declaration and validation facility for the Java application developer, as well as a constraint metadata repository and query API.
```
gsmet: This one fixes references to JavaBean. It should be squashed as a fixup commit.

sources/whatsnew.asciidoc
```diff
@@ -10,8 +10,8 @@
 [[whatsnew-20]]
 === What's new in 2.0
 
-The main contribution of {spec-name-bv} 2.0 is leveraging the new language features and API additions of Java 8 for the purposes of validation.
-Java 8 or later is required to use {spec-name-bv} 2.0.
+The main contribution of Bean Validation 2.0 is leveraging the new language features and API additions of Java 8 for the purposes of validation.
+Java 8 or later is required to use Bean Validation 2.0.
```
gsmet: I don't think it's a good idea to mention Jakarta in the changelog, in the context of BV 2.0 and BV 1.1.

I could envision that we do it for 2.0 (even if it's rewriting history) but clearly not for 1.1.

sources/introduction.asciidoc
```diff
@@ -17,7 +17,7 @@ It also offers method and constructor validation facilities to ensure constraint
 This work has been conducted as part of the {spec-name-bv} under the link:https://jakarta.ee/about/jesp/[Jakarta EE Specification Process] and formerly JSR 380, 349 and 303 under the Java Community Process Program.
 This specification is the result of the collaborative work of the members of the Expert Groups and the community at large.
 
-The following persons have actively contributed to {spec-name-bv} 2.0 as members of the JSR 380 expert group and the community at large in alphabetical order:
+The following persons have actively contributed to Bean Validation 2.0 as members of the JSR 380 expert group and the community at large in alphabetical order:
```
gsmet: I think these mentions should stay as is.

sources/constraint-declaration-validation.asciidoc
```diff
@@ -138,7 +138,7 @@ A future revision of this specification might define support for such usages of
 ----
 public class User {
 
-    // preferred style as of Jakarta {spec-name-bv} 2.0
+    // preferred style as of Jakarta Bean Validation 2.0
```
gsmet: Note as for this one and the one above, we could enable attribute replacement in the code instead.

For the XSD file, we really need to hardcode the value as it's supposed to be verbatim the content of the files.

sources/introduction.asciidoc
emmanuelbernard: oh man, I looked at this one for ever, I knew I had made a mistake like that but could not spot it. Drove  me nuts.

sources/whatsnew.asciidoc
emmanuelbernard: OK as I am not sure, I've created a  separate commit (like yours) but added a note explaining why we don't change,

sources/constraint-declaration-validation.asciidoc
emmanuelbernard: This is what I had intended to do. It just avoided my attention

PR #255, opened 2020-02-03
Fix protocol (http->https) for Maven Central

pzygielo:
```
501 HTTPS Required.
Use https://repo.maven.apache.org/maven2/
More information at https://links.sonatype.com/central/501-https-required
```


gsmet:
Missed this one as I had everything locally. Good catch.

Issue #274, opened 2024-01-26
We need to modernize the specification doc build


starksm64:
The current spec document build is based on ant and the asciidoctor ant plugin. The ant plugin has not been updated in 5 years, and we are using an even older version of it. Even just updating to the 5 year old version of the ant plugin causes the generation to fail because the custom Hibernate tree/block asciidoc plugins are apparently incompatible with that version.

We should look at switching the generation to the maven plugin, but this will probably still require updates to the hibernate-asciidoctor-extensions.

PR #276, opened 2024-02-02
Maven build

mswatosh:
This should be a good base for #274. It works with no changes to the asciidoc files.

Builds the spec:
- html
- pdf
- docbook (with tck block processing)

Not all the functionality of the build.xml is ported, including things like the CI variables. I'm not sure what all should be ported vs what can be left in the build.xml. 

The only major piece missing as far as I can tell is building the preprocessed adoc, which works, but will require a new version of hibernate-asciidoctor-extensions with this PR included:
https://github.com/hibernate/hibernate-asciidoctor-extensions/pull/51

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-!# Jakarta Validation specification
+# Jakarta Validation specification
 
 This repository contains the Jakarta Validation specification. 
 For more information on Jakarta Validation and the work in progress,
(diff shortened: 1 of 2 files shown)


starksm64:
So I tried building your branch with 2.0.1-SNAPSHOT build of the hibernate-asciidoctor-extensions and the differences in the html versions are insignificant. I'll merge this as it is an improvement.

PR #287, opened 2025-05-16
Fix incorrect example for `@Size`

DavideD:
Fix jakartaee/validation#259


DavideD:
ECA signed

marko-bekhta (approved):
Thanks Davide! 🎉

PR #288, opened 2025-05-19
Undefined behaviour when @Valid is applied twice

DavideD:
```java
@Valid
private final List<@Valid> VisitorWithGroups> visitors;
```

It wasn't clear whether the elements of the visitors list should be validated once or twice. The specification now states the the behaviour is undefined and left to the specific implementation.

See https://hibernate.atlassian.net/browse/BVAL-709


sources/constraint-declaration-validation.asciidoc
```diff
@@ -123,7 +123,9 @@ See <<valueextractordefinition-builtinvalueextractors>> for the complete list of
 [tck-testable]#`@Valid` also allows the validation of the elements of nested generic containers. `@Valid` must be put to a type argument of that nested container type in order to trigger validation of the elements of all the nested containers.#
 
 For a given container, the `@Valid` annotation should either be put to the container itself _or_ to the type argument(s) of the container,
-but not both (in order to prevent the container elements from being validated twice).
+but not both.
+[tck-not-testable]#This specification dosen't define the behaviour for having the `@Valid` annotation on the container _and_ on the type argument(s) of the container.
```
pzygielo, line 127: _dosen't_ -> _doesn't_ or even _does not_.

DavideD: fixed, thanks

DavideD:
I removed the code example and updated the commit message, just in case

DavideD:
ECA signed

